PR #30, opened 2014-12-09
Add legacy MD5 Minter

no-reply:
Submitting this for discussion.

Interested in ideas about more robust testing of edge cases for existing ID's. This needs to be able to replicate https://github.com/dpla/ingestion/blob/develop/lib/akamod/select-id.py#L55

### DIFF
--- a/lib/krikri.rb
+++ b/lib/krikri.rb
@@ -10,9 +10,4 @@ module Krikri
   autoload :OaiDcParser,    'krikri/parsers/oai_dc_parser'
   autoload :JsonParser,     'krikri/parsers/json_parser'
   autoload :ModsParser,     'krikri/parsers/mods_parser'
-
-  # Enrichments
-  autoload :Enrichment,       'krikri/enrichment'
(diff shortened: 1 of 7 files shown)


no-reply:
Discussion about OriginalRecord ids happened in slack, running from here: https://dpla.slack.com/archives/tech/p1421080663000069

no-reply:
This is updated to serve as the main ID minter for OriginalRecords.

Given the correct provider prefix, this will generate a legacy compatible ID for most OriginalRecords, but it does not attempt to replicate IDs from the legacy edge cases (primo harvesters).

anarchivist:
I think this looks good to me, and pending any other feedback, this looks good to :squash:

markbreedlove:
That's great; I'm all for it.

no-reply:
:squash:ed. Leaving the Harvester refactor in a separate commit, since it's an essentially unrelated change.

anarchivist:
:key: :passport_control:

PR #39, opened 2014-12-22
Feature/abstract enrichment

no-reply:
This requires documentation.  There are also multiple "complex enrichments" in the works. I expect at least one to get added.

Note that shared examples for complex enrichments don't actually run in the current code base.

### DIFF
--- a/lib/krikri/enrichment.rb
+++ b/lib/krikri/enrichment.rb
@@ -1,0 +1,148 @@
+module Krikri
+  ##
+  # Mixin module for enriching a set of input_fields and setting the resulting
+  # values to a set of output fields.
+  module Enrichment
+    extend SoftwareAgent
+
+    ##
+    # The main enrichment method; passes specified input fields to
+    # #enrich_values, which must return an array of values with length equal to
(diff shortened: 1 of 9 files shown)


lib/krikri/enrichment.rb
```diff
@@ -0,0 +1,148 @@
+module Krikri
+  ##
+  # Mixin module for enriching a set of input_fields and setting the resulting
+  # values to a set of output fields.
+  module Enrichment
+    extend SoftwareAgent
+
+    ##
+    # The main enrichment method; passes specified input fields to
+    # #enrich_values, which must return an array of values with length equal to
```
anarchivist: How does this impact enrichment stories like [#84945216](https://www.pivotaltracker.com/story/show/84945216) (deduplication) and [#83401638](https://www.pivotaltracker.com/story/show/83401638) (separate at semicolons)?

anarchivist: I see, never mind. This is a case of `Enrichment` vs `FieldEnrichment`.

anarchivist:
OK, I think this looks great and ready to squash.

PR #45, opened 2015-01-06
Change default MIME Type for Original Records

no-reply:
Marmotta assigns file extensions based on content type. The octet stream MIME Type causes OriginalRecords to have the `.bin` extension. This changes the default MIME to `text/xml` so the default extension will be `.xml`.

Note that Marmotta infers that `application/xml`, `text/plain`, and some other types are RDF Sources.

An unrelated minor fix I spotted while working on this is included as a separate commit.  I can rebase this into another branch/PR if desired.

### DIFF
--- a/spec/models/original_record_spec.rb
+++ b/spec/models/original_record_spec.rb
@@ -147,9 +147,10 @@ describe Krikri::OriginalRecord do
 
     it 'reloads content_type' do
       subject.save
+      ctype = subject.content_type
       subject.content_type = 'application/xml'
       subject.reload
-      expect(subject.content_type).to eq 'application/octet-stream'
+      expect(subject.content_type).to eq ctype
     end
   end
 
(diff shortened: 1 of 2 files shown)


spec/models/original_record_spec.rb
```diff
@@ -147,9 +147,10 @@
 
     it 'reloads content_type' do
```
markbreedlove: Since it's not being saved after line 151, can we not assume that subject.content_type is going to be the same after reload?

no-reply: Exactly, that's what we're testing.  We want the `content_type` to sync from Marmotta's entry on reload.  The behavior of the test doesn't change here, it's just not hard coded anymore.

anarchivist:
:+1: but with a quick question: should instances of `Krikri::Harvester` set `content_type` before saving?

no-reply:
Probably. This is now called out in a task on the CDL Harvester story (which is likely to be the first non-XML harvester): https://www.pivotaltracker.com/story/show/84398770

PR #46, opened 2015-01-06
Add a registry of harvesters

markbreedlove:
Add a registry of harvesters, `Krikri::Harvester::Registry`.   This will be used when we enqueue jobs, or in the eventual user interface, in order to retrieve a harvester class given a token (e.g. get an OAIHarvester class for "OAI") or to retrieve a list of available harvester tokens for the purpose building menus.

A future Event will look up the harvester token from the HarvestSource for a given institution, and call `#enqueue` on the relevant Harvester to put it in the queue.

Coming soon in another PR:  make a harvester report on the options that it will need when instantiated.  This will be necessary in the future for any client that needs to create harvest jobs with the right options for `enqueue`.


lib/krikri/harvester.rb
```diff
@@ -24,7 +24,14 @@ module Krikri
...
   #      Runs a full harvest, inserting Original Records into the database,
   #      given the options passed to #initialize.
-  class Harvester
+  #
+  # When including, add a call to Krikri::Harvester::Registry.register() to
+  # put it in the registry so that it can be looked up.
+  # See lib/krikri/engine.rb.
+  # @see Krikri::Engine
+  #
+  module Harvester
```
no-reply: :+1:

lib/krikri/harvester.rb
```diff
@@ -77,5 +84,40 @@ def run
       true
     end
 
+    ##
+    # An application-wide registry of defined Harvesters.
+    class Registry
```
no-reply: What are the differences between this and the mapping registry? How much of this, and the related tests, is duplicated from there?

spec/lib/krikri/harvester_spec.rb
```diff
@@ -1,19 +1,61 @@
...
+  # Krikri::Harvester.
+  let(:klass) { Class.new }
+  subject { klass.include(Krikri::Harvester).new }
 
   context 'with record_ids implemented' do
     before do
-      described_class.send(:define_method, :record_ids) do
-        [1, 2, 3, 4, 5]
-      end
+      expect(subject).to receive(:record_ids).and_return([1, 2, 3, 4, 5])
```
no-reply: This is a significant improvement. :+1:

Maybe it's just an `allow` rather than an `expect`, though.  Do all possible tests in this context *need* to call `#record_ids`?

no-reply:
Looks good broadly. If you think it can't/shouldn't be refactored to share code with the mapping registry, then there aren't any major changes I would propose.

lib/krikri/harvester.rb
markbreedlove: I could move this to `Krikri::Registry`, and have the harvester and mapper registries extend it.
There's not much difference, other than that I added a `#keys` method for the harvester one.

no-reply: I'm in favor of `Krikri::Registry`.  Seems like `#keys` would be as relevant to the existing registry as this one, yes?

spec/lib/krikri/harvester_spec.rb
markbreedlove: I think that this was only to support the 'knows its record count' example.

My assumption, from the way it had been written, was that `#count` invokes `#record_ids`.  I was probably also going by the `context`, which says, "with record_ids implemented."  I didn't know about `about`, though, and I see now how it's softer than `expect`.  I could change it.  However, I wonder if doing so would let us put tests in this context that shouldn't be there.  If they never called `record_ids` then should they be in this `context` block?  Maybe `expect` would flag that.  I'm not super invested in `expect`, whatever the case.

no-reply: I'm not super invested, either.  Your argument that this context should only contain specs that call `record_ids` sounds good. 

The alternative perspective is that this `expect` over specifies implementation details.  Specifically: `allow` means that when `#record_ids` is implemented `#count` should work; `expect` means that `#count` should work *and* call `#record_ids`.

Either works for now, but I suspect `allow` will prove more durable.

markbreedlove: OK, with that in mind, I think it's a good idea to switch to `allow`.

lib/krikri/harvester.rb
markbreedlove: Yes, I think `#keys` would be generally useful.  I'm making these changes now.

no-reply:
This looks ready to :squash: to me.

lib/krikri/registry.rb
```diff
@@ -0,0 +1,45 @@
+module Krikri
+
+  ##
+  # Provide a registry of defined items that can be looked up by a token
```
anarchivist: This may be nitpicky as well, but does this have to be a string? IIRC, our examples of adding a new mapping to the registry use symbols and not strings. (See also [here](https://github.com/dpla/KriKri/pull/46/files#diff-b31c2deb469ce65b5f99d9b33041a7c9R56).)

lib/krikri/engine.rb
```diff
@@ -51,5 +51,10 @@ class Engine < ::Rails::Engine
       end
     end
 
+    initializer :register_harvesters do
+      Krikri::Harvester::Registry
+        .register('OAI', Krikri::Harvesters::OAIHarvester)
```
anarchivist: Should this be `Krikri::Harvester::Registry.(:oai, Krikri::Harvesters::OAIHarvester)`? (See my comment [here](https://github.com/dpla/KriKri/pull/46/files#r22553261).)

lib/krikri/registry.rb
anarchivist: IOW, do we want to prefer symbols to strings, and do we need to do any string-to-symbol conversion here to assure appropriate resolution?

spec/lib/krikri/harvesters/oai_harvester_spec.rb
```diff
@@ -229,3 +229,14 @@
     it_behaves_like 'a harvester'
   end
 end
+
+
+describe Krikri::Harvester::Registry do
+  describe '#registered?' do
+    it 'knows OAIHarvester is registered' do
+      # It should have been registered by the engine initializer, engine.rb.
+      expect(described_class.registered?('OAI')).to be true
```
anarchivist: See [here](https://github.com/dpla/KriKri/pull/46/files#r22553261) and [here](https://github.com/dpla/KriKri/pull/46/files#r22553388).

spec/lib/krikri/registry_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'spec_helper'
+
+
+describe Krikri::Registry do
+
+  before(:all) do
+    described_class.register('mock', Class.new)
```
anarchivist: See [here](https://github.com/dpla/KriKri/pull/46/files#r22553261) and [here](https://github.com/dpla/KriKri/pull/46/files#r22553388).

lib/krikri/registry.rb
markbreedlove: A string is going to come from the database, when you retrieve the HarvestSource record for an institution (harvest_sources.source_type).  It just seemed more direct to expect a string for that reason.  I know that calling String.to_sym would be pretty trivial but it would be an extra step, possibly confusing to someone.

markbreedlove: Also, the other thing is you'll want at some point to list the keys of the registry to make a menu in some u.i. of possible harvester types and then you'd have to convert them to strings.

anarchivist: Re: your first point, I agree with @no-reply's comment in Slack: 

```
< tom> ah, seeing the database issue, i'm now more in favor of #to_sym
``` 

Regarding your second point (UI issues of harvester types), it seems like that ideally we'd best be dealing with that through defining locale files for `I18n`.

anarchivist: See also this StackOverflow answer: [When to use symbols versus strings in Ruby?](http://stackoverflow.com/a/16621092/1277323)

markbreedlove: Thanks, I recall the efficiency argument but just wasn't sure about converting back and forth.
I'll change it to a symbol if you think that's best.

anarchivist:
Assuming the tests pass on Travis I think this is ready to squash.

PR #47, opened 2015-01-06
Set Jettywrapper URL and Marmotta Home in host

no-reply:
This allows a host application to run jetty rake tasks successfully.


anarchivist:
:+1:

PR #48, opened 2015-01-07
Add discovery of Harvester options

markbreedlove:
This adds an `opts` field to `HarvestSource` for storing serialized options that get passed to the harvester.  For example, a list of sets for an OAI harvester.

`Krikri::Harvester.expected_opts` has been added to help client code know what options to get from the user to store in `HarvestSource`.  See the comment below for that method.

OAIHarvester has also been changed to take 'uri' as an option instead of 'endpoint', to bring it inline with the HarvestSource model, which has a 'uri' field, not an 'endoint' field.

The next step, assuming this goes over well, is to make an Event that can read the `opts` in a HarvestSource and determine the right strategy for enqueueing jobs for the relevant harvester, e.g. do one harvest job for the whole harvest, or iterate over sets, using the same URI each time, etc.

### DIFF
--- a/db/migrate/20150106164218_add_opts_to_krikri_harvest_sources.rb
+++ b/db/migrate/20150106164218_add_opts_to_krikri_harvest_sources.rb
@@ -1,0 +1,5 @@
+class AddOptsToKrikriHarvestSources < ActiveRecord::Migration
+  def change
+    add_column :krikri_harvest_sources, :opts, :text
+  end
+end
--- a/lib/krikri/harvester.rb
+++ b/lib/krikri/harvester.rb
@@ -88,5 +88,42 @@ module Krikri
     # An application-wide registry of defined Harvesters.
     Registry = Class.new(Krikri::Registry)
 
+    ##
+    # @abstract Return initialization options for the harvester.
+    #   The harvester will expect to receive these upon instantiation (in the
+    #   opts argument to #initialize), in the form:
+    #   {
+    #      key: <symbol for this harvester>,
+    #      opts: {
+    #        option_name: {type: :type, required: <boolean>,
+    #                      multiple_ok: <boolean, default false>}
+    #      }
+    #   }
+    #   ... where type could be :uri, :string, :int, etc.
+    #   ... and multiple_ok means whether it's allowed to be an array
+    #   ... for example, for OAI this might be:
+    #   {key: :oai,
+    #    set: {type: :string, required: false, multiple_ok: true},
+    #    metadata_prefix: {type: string, required: true}}
+    #
+    # @todo The actual type token values and how they'll be used is to be
+    #   determined, but something should exist for providing validation
+    #   guidelines to a client so it doesn't have to have inside knowledge of
+    #   the harvester's code.
+    #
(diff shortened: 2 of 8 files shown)


no-reply:
I'm good with this in general (I haven't read the code closely yet, but it looks good at a glance).

```OAIHarvester has also been changed to take 'uri' as an option instead of 'endpoint', to bring it inline with the HarvestSource model, which has a 'uri' field, not an 'endoint' field.```

I'm wondering about the importance of this. It suggests that you are thinking to couple the `HarvestSource` fields and their names pretty tightly to `Harvester`.  My philosophy would be to keep the dependency loose and strictly a dependency of `HarvestSource` on some portion of the `Harvester` interface.

Maybe rather than trying to sync the field names, an approach that generates the `HarvestSource` form from the options presented by the relevant `Harvester` is a good idea?

(I could be reading too much into this change...)

markbreedlove:
I just figured that `uri` is / has been used more commonly elsewhere, and I didn't want anyone to look at this argument as being different in some way.  It's mostly consistency for its own sake here.  The term `uri` would be appropriate for a file-based harvester as well as a networked one, and we can count on every harvester having this argument.  We would therefore avoid momentary confusion with the OAIHarvester if someone used to reading `uri` has to look at `endpoint` and wonder why it's different.

BTW, the trick really is the loose coupling, and why I'm taking the creation of the ingestion event slowly.  There are all of these different harvesters that will work in different ways, and HarvestSource has to be generic, and the `opts` value has to be considered carefully.

As you suggest, yes, I think the HarvestSource should be filled in based on the fields that the particular Harvester says it wants.  I've added that `opts` field for anything that's not `uri`, just because `uri` is the one thing that I think is universal for all harvesters.  I think that `opts` will be a JSON string.

anarchivist:
I think this looks good as well, pending any other feedback from @no-reply. I think my one point for discussion, picking up on Tom's earlier comment, would be to think about how to set up the `opts` for harvester-type-specific `Krikri::Harvester` subclasses. For example, if there are notions that just relate to `Krikri::Harvesters::OAIHarvester`, would/could those be relegated to an `oai` key in the hash?

markbreedlove:
Perhaps, yes, I hadn't thought of an `oai` key per se, but I have a branch I'm about to push that shows part of the approach I'm thinking of for harvester-specific details.

no-reply:
This looks good to me. With the clarification in @markbreedlove's comment, it looks like our thinking about this is aligned.

I'll look at the newer branch in a bit; pending anything coming out of that, I don't have a strong opinion about the structure of the ops hash.  @anarchivist can you say something about the benefit of a top-level `oai` key vs. a flatter structure?

anarchivist:
@no-reply I don't have a strong opinion about the structure per se, but having some mechanism for namespacing these harvester-specific options might be preferable.

markbreedlove:
I'll see how it works with an `oai` key.  I think that would make it more clear.

markbreedlove:
OK, i've added that `oai` key.

no-reply:
This looks good to me. Squashy, squashy?

PR #49, opened 2015-01-07
Ldp connection refused

no-reply:
Connection Refused errors on LDP requests were being caught as ClientErrors, but do not have a `#response`, so were throwing no method on `nil` errors instead of something informative. This ensures that a useful error is always thrown.

I ran into this problem when messing around with `heidrun`.  I wrote a quick patch, and a test.  Maybe the test is a bit naive, since it relies on no activity on port 9999 on the host machine.  I couldn't find a way to get WebMock to mimic Connection Refused (with fairly minimal searching, I admit).

### DIFF
--- a/lib/krikri/ldp/resource.rb
+++ b/lib/krikri/ldp/resource.rb
@@ -69,7 +69,7 @@ module Krikri::LDP
     rescue Faraday::ResourceNotFound
       false
     rescue Faraday::ClientError => e
-      return false if e.response[:status] == 410
+      return false if !e.response.nil? && e.response[:status] == 410
       raise e
     end
     alias_method :exist?, :exists?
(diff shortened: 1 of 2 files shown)


markbreedlove:
I wonder what would happen if you tried port 0.  I think that's "reserved" and nothing actually uses it.

markbreedlove:
BTW, port 9999 is for something called "distinct": http://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=9999

Maybe you could also use one of the "unassigned" ports, like 4, for example: http://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml

anarchivist:
I'm :+1: on this with @markbreedlove's suggestion.

no-reply:
0 seems to work. :squash:

PR #51, opened 2015-01-07
Remove Faraday User-Agent headers from mocks

markbreedlove:
This should clear up errors related to our having upgraded Faraday.


no-reply:
Looks good. When the build finishes, I'll :merge:.

PR #52, opened 2015-01-08
Add codeclimate test coverage reporting

no-reply:
Assuming we want this, it should work.  Test coverage reporting isn't perfect (it will never tell you whether you're testing effectively), but it might help us spot problem areas.

http://blog.codeclimate.com/blog/2013/09/11/test-coverage-and-code-quality-better-together/


anarchivist:
:partly_sunny: :shipit:

PR #54, opened 2015-01-09
Update DPLA MAP dependency to pre.7

no-reply:
Updates the solr schema, makes some changes to property access in tests.

More importantly, pins this version tightly to `pre.7`. Previous `pre` versions of MAP haven't been compatible when updating, so it seems like a good idea to be pessimistic until we can release 4.0.0.0.


anarchivist:
:question: :squash: :grey_question:

PR #55, opened 2015-01-11
Normalize OAI output

no-reply:
Ensures that OAI `#records` and `#get_record` return the same result (content) for a given identifier.

Unfortunately, `OAI:Record` hides its raw XML, so it is necessary to rebuild the record from its parts.  The alternative is to monkeypatch an `element` attribute.

```ruby
class OAI::Record
  attr_reader :element

  def initialize(element)
    @element = element
    @status = get_attribute(element, 'status')
    @identifier = xpath(element, './/identifier')
    @datestamp = xpath(element, './/datestamp')
    @set_spec = xpath_all(element, './/setSpec')
  end
end
```
https://github.com/code4lib/ruby-oai/blob/master/lib/oai/client/record.rb

This ignores OAI provenance content from `<about>...</about>`.

### DIFF
--- a/lib/krikri/harvesters/oai_harvester.rb
+++ b/lib/krikri/harvesters/oai_harvester.rb
@@ -51,7 +51,7 @@ module Krikri::Harvesters
     def records(opts = {})
       opts = opts.merge(@opts)
       client.list_records(opts).full.lazy.flat_map do |rec|
-        Krikri::OriginalRecord.build(rec.header.identifier, rec.metadata.to_s)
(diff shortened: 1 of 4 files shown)


anarchivist:
I think this is good to go, but as discussed in Slack, we may want to consider sending a patch to `code4lb/ruby-oai`.

no-reply:
I submitted an issue https://github.com/code4lib/ruby-oai/issues/41

no-reply:
:squash:ed

PR #56, opened 2015-01-12
Refactor SoftwareAgent tests for shared examples

no-reply:
The previous SoftwareAgent tests tested a representative example, but could not be applied to new inheriting classes. The shared examples allow the behavior of individual SoftwareAgents to be tested.


no-reply:
This is all prep work for _Map Harvested Records_ https://www.pivotaltracker.com/story/show/82567674. 

I'm closing it so it can be included in that (forthcoming) PR.

PR #57, opened 2015-01-12
Update dpla-map to 4.0.0.0-pre8; ignore Redis backup file

anarchivist:


no-reply:
:+1: Will merge when :travis: is happy.

PR #58, opened 2015-01-15
Allow Registry to force-register items

markbreedlove:
Add #register! in addition to existing #register

This accompanies a forthcoming PR in Heidrun that will allow Heidrun Resque workers to be started.


no-reply:
:rocket: :squash:

PR #60, opened 2015-01-19
Run Mappings Against a Harvest

no-reply:
Adds the ability to run a mapping on a whole harvest and save the resulting records over LDP.  Significant refactors of Agents and Jobs are included, as well as LDP RDFSource integration for ActiveTriples.

 - `Krikri::Activity`: Now contains the logic to build activities from its class name and opts, as well as an `#rdf_subject` so it can be referred to in LDP Graphs.  We could add capability for it to serialize more of its data as RDF, but no use cas stands out yet.

 - `Krikri::OriginalRecord`: Now has metadata, backed by `Krikri::OriginalRecordMetadata` (an `ActiveTriples::Resource` / `Krikri::Ldp::RdfSource`.  This allows us to store the knowledge that a given `Krikri::Activity` (harvest) generated the record.

 - `DPLA::MAP::Aggregation`: Is monkey patched to include `Krikri::Ldp::RdfSource`, allowing MAP records to be persisted/retrieved over LDP.  It also gets the ability to mint its own identifier.

 - `Krikri::Job`: Gets more of the core logic of jobs moved out of its subclasses and toward the base class.  It's possible that all jobs will end up being the same, just with different queue names. For now, Jobs implement `#run` to complete the interface and actually process the job.

- `Krikri::Ldp::Resource`: Now validates that the request URI's target the configured marmotta endpoint.

- `Krikri::Mapper`: Now includes a `Krikri::Mapper::Agent` as the responsible `SoftwareAgent` for running a mapping job.

- `Krikri::Mapping`: Now accepts a parser at time of definition.  When records are passed in, it uses this parser to interpret them, unless they are already parsed.

- `Krikri::ProvenanceQueryClient`: This is an initial attempt at a system for querying Marmotta for items by the `Activity` that created them.  I hope that it will grow over time. For now it implements a single query and ties in (half heartedly) with `SPARQL::Client`.

- Several `_spec` files are refactored or moved to use shared examples.


lib/krikri/mapper.rb
```diff
@@ -62,6 +62,37 @@ def map(name, records)
...
+      attr_reader :name, :generator_uri
+
+      def initialize(opts = {})
+        @name = opts.fetch(:name).to_sym
+        @generator_uri = RDF::URI(opts.fetch(:generator_uri))
+      end
+
+      def run(activity_uri = nil)
+        Krikri::Mapper.map(name, records).each do |rec|
+          rec.set_subject!(SecureRandom.hex)
```
no-reply: I want to have a conversation with the team about how to handle identifiers and which component should have the responsibility of assigning them.

I have considered doing this in the mapping (via the existing `#uri` DSL method), but don't have a plan for handling mappings that fail to set a URI.  A URI is necessary to save a `DPLA::MAP::Aggregation` as an LDP resource, so it's insufficient to allow objects w/out URI's to remain blank nodes.

lib/krikri/job.rb
```diff
@@ -2,12 +2,32 @@ module Krikri
...
   class Job
     @queue = nil
     ##
-    # @abstract Perform the job; called by the queue system
-    def self.perform(*)
-      fail NotImplementedError
+    # Perform the job.
+    def self.perform(activity_id)
+      activity = Krikri::Activity.find(activity_id)
+      activity.run { |agent, activity_uri| run(agent, activity_uri) }
```
markbreedlove: This is good to see how to do this, along with the revised Activity#run.  The abstraction is good.

no-reply: I've thought about removing the `Job#run` middleman, since both existing implementations are effectively the same.  I'm sort of waiting to see what happens with enrichment queuing, but I suspect all relevant arguments want to be passed to the agent as options, yes?

app/models/krikri/activity.rb
```diff
@@ -27,13 +26,34 @@ def set_end_time
...
       end
     end
 
+    ##
+    # Instantiates and returns an instance of the Agent class with the values in
+    # opts.
+    #
+    # @return [Agent] an instance of the class stored in Agent
+    def agent_instance
+      @angent_instance ||= agent.constantize.new(parsed_opts)
```
markbreedlove: The only thing that's confusing me so far about this refactor is the scope of `parsed_opts` and `opts`.  Where are they defined?

markbreedlove: Uh, I mean actually `opts` ... in `#parsed_opts`

no-reply: It's an attribute of the model.  It's defined via ActiveRecord magic in the database table.

spec/ldp/rdf_source_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'spec_helper'
+
+describe Krikri::LDP::RdfSource do
```
no-reply: This file should be removed. These tests are handled in shared examples, here: https://github.com/dpla/KriKri/pull/60/files#diff-dadee38ed9146031364c384d8615bd03

app/models/krikri/activity.rb
anarchivist: Should this be `@agent_instance` and not `@angent_instance`?

app/models/krikri/activity.rb
```diff
@@ -27,13 +36,34 @@ def set_end_time
...
+    end
+
+    def parsed_opts
+      JSON.parse(opts, symbolize_names: true)
+    end
+
+    def rdf_subject
+      RDF::URI(File.join(Krikri::Settings['marmotta']['ldp'],
```
anarchivist: Should we be using `RDF::URI#/` here instead of `File.join`?

no-reply: Absolutely! Good call.

app/models/krikri/activity.rb
no-reply: Typo. Will fix. Good catch.

app/models/krikri/original_record_metadata.rb
```diff
@@ -0,0 +1,14 @@
+module Krikri
+  ##
+  # Descriptive Metadata for OriginalRecord
+  class OriginalRecordMetadata < ActiveTriples::Resource
```
anarchivist: Might be worth a chat with Gretchen and Amy if we want to call this out in a future version of the DPLA MAP. (Maybe 4.1, just to make the near-term too confusing.)

no-reply: I'm :+1: on that.

lib/krikri/engine.rb
```diff
@@ -56,5 +56,32 @@ class Engine < ::Rails::Engine
...
 
+    initializer :rdf_repository do
+      Krikri::Repository =
+        RDF::Marmotta.new(Krikri::Settings['marmotta']['base'])
+    end
+
+    initializer :aggregation do
+      DPLA::MAP::Aggregation.class_eval do
+        include Krikri::LDP::RdfSource
+        configure :base_uri => Krikri::Settings['marmotta']['item_container']
```
anarchivist: Curious about what precipitated this. Don't have a strong feeling about it though.

no-reply: Which part in particular?

It's necessary to reopen the class to `include Krikri::LDP::RdfSource`, which isn't necessarily desirable in the broader `dpla-map` context.  The rest of this could conceivably be moved upstream.

anarchivist: Sorry, I meant specifically the addition of an LDP container for the items.

no-reply: Every `Resource` must go in a `Container`. Though marmotta will create an `ldp:BasicContainer` for the items, it's effectively just a base path.

lib/krikri/harvester.rb
```diff
@@ -100,9 +100,9 @@ def get_record(_)
     # This should be idempotent so it can be safely retried on errors.
     #
     # @return [Boolean]
-    def run
+    def run(activity_uri = nil)
       log :info, 'harvest is running'
-      records.each(&:save)
+      records.each { |rec| rec.save(activity_uri) }
```
anarchivist: Maybe I'm asking this in the wrong context, but does this allow us to run partial harvests?

no-reply: Not through `#run` (and therefore not through the `Job` queue). You could do it various other ways, and still save with an `activity_uri`.

As you seem to be pointing out, this change doesn't impact that issue either way.

anarchivist: OK, yep, I think that's fine. Just wrapping my head around things.

lib/krikri/mapper.rb
```diff
@@ -62,6 +63,37 @@ def map(name, records)
...
+      def run(activity_uri = nil)
+        Krikri::Mapper.map(name, records).each do |rec|
+          rec.mint_id! if rec.node?
+          rec << RDF::Statement(rec, RDF::PROV.wasGeneratedBy, activity_uri) if
+            activity_uri
+          rec.save
+        end
+      end
+
+      def records
```
anarchivist: Looking at your concern raised in the specs for this and `Krikri::ProvenanceQueryClient`, maybe  we should add a method that returns a new `Krikri::ProvenanceQueryClient` instance. Then you can ensure that it responds to `execute` as a subclass of `SPARQL::Client` in its own spec, maybe like [here](https://github.com/ruby-rdf/sparql-client/blob/develop/spec/client_spec.rb#L179-L188).

spec/lib/krikri/mapper_agent_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    let(:record_double) { double(Krikri::OriginalRecord) }
+
+    it 'returns a lazy enum' do
+      expect(subject.records).to be_a Enumerator::Lazy
+    end
+
+    it 'gets records for generated by harvest_activity' do
+      expect_any_instance_of(Krikri::ProvenanceQueryClient)
+        .to receive(:find_by_activity).with(generator_uri).and_return(query)
+      expect(query).to receive(:execute).and_return(solutions)
```
anarchivist: Hmmm. Lines 60-62 (although maybe not 62) feel like their integration tests for `RDF::Query`, `RDF::Query::Solutions`, and `RDF::Query::Solution`. Are they necessary here?

Alternately, do we just need a factory to return `RDF::Query::Solutions` for a provenance query?

no-reply: These lines mock out the modules you reference exactly to avoid being an integration test.  The problem here is that there's a long list of dependencies that any user of `ProvenanceQueryClient` takes on board.  

I think the solution must be to have that module insulite its clients from `RDF::Query`, `RDF::Query::Solutions` and `RDF::Query::Solution`?

anarchivist:
Looking good. Sounds like @no-reply is going to take a bit of a look at refactoring `Krikri::Mapper::Agent#records` and `Krikri::ProvenanceQueryClient`.

Also, as noted in Slack discussion

```
< mark> [12:53] mmm maybe we just need a generic _module_ for our canned queries rather than something that inherits from `SPARQL::Client`
< tom> [12:53] that's what i'm starting to think, too
< mark> [12:53] because i'm not seeing why we need to subclass it
< tom> [12:53] the inheritance is maybe too clever for its own good
```

app/models/krikri/activity.rb
```diff
@@ -8,6 +8,15 @@ module Krikri
   # it is -- harvest, enrichment, etc.
   #
   class Activity < ActiveRecord::Base
+    # @!attribute agent
+    #    @return [Class] a Krikri::SoftwareAgent responsible for the activity
```
AudreyAltman: When you say you are returning `[Class]` is this a Class Object or a String that represents a Class Object?

spec/models/activity_spec.rb
```diff
@@ -45,4 +54,26 @@
...
   end
 
+  describe '#agent_instance' do
+    it 'returns an instance of the agent class' do
+      expect(subject.agent_instance)
+        .to be_an_instance_of(subject.agent.constantize)
+    end
+
+    it 'returns the same instance for successive calls' do
+      expect(subject.agent_instance).to eq subject.agent_instance
```
AudreyAltman: I'm not sure of exactly what you're trying to test here, but are you sure you want `eq`?  Are you testing whether it returns the same object or equivalent objects?  Your `it` statement makes it sound like you want the same object, in which case I believe you should use `be` instead of `eq`.

app/models/krikri/original_record.rb
```diff
@@ -8,13 +8,17 @@ class OriginalRecord
     attr_writer :content_type
 
     def initialize(identifier)
+      raise ArgumentError, "#{identifier} is an invalid local name" if
+        identifier.include?('/')
       @local_name = identifier
     end
```
AudreyAltman: It might be nice to include a comment explaining when to instantiate an instance OriginalRecord vs. when to interact with OriginalRecord as a class.  It was very helpful when you explained it to me.

lib/krikri/job.rb
```diff
@@ -2,12 +2,32 @@ module Krikri
   ##
   # Generic Job class the gets extended by specific types of Jobs;
```
AudreyAltman: Typo in line above, remove "the".

spec/models/activity_spec.rb
no-reply: You have it right. Nice catch!

lib/krikri/ldp/rdf_source.rb
```diff
@@ -0,0 +1,29 @@
...
+      result = super(dump(:ttl))
+      get({}, true)
+      result
+    end
+
+    def get(*args)
+      result = super
+      reload_ldp
+      result
+    end
```
AudreyAltman: It would be super awesome if you could include a comment saying what object type a call to `get` will return (including the possibility of errors being thrown).

no-reply: Would a `@see Krikri::LDP::Resource#get` serve, here? https://github.com/dpla/KriKri/blob/develop/lib/krikri/ldp/resource.rb#L56

AudreyAltman: :+1:

no-reply:
I'm rebasing this branch.

app/models/krikri/original_record.rb
no-reply: Comments added.

AudreyAltman:
Aside from my small suggestions about documentation and `activity_spec`, this all looks good to me.

no-reply:
Great. I think all suggestions are resolved. I'll set about squashing this (that may be a task!), in a little bit.

no-reply:
Rebased, :squash:ed, and force pushed.

PR #61, opened 2015-01-20
DSL Methods & Resources

no-reply:
Adds functionality to the mapping DSL for creating instances of a resource class from each of a given parser node's values and for accessing parsed values in a more human readable fashion.

Creates a `RecordProxy` in the mapping DSL context. The special DSL method `#record` now returns an instance of this class.  The class accepts a call chain, with method names, their arguments (and blocks to pass through) which will be called against the parsed record (via `ValueArray`) when the record is processed.  `RecordProxy` mimics `Proc` in that it has `#call` and `#arity` methods.  This allows the rest of the DSL's interface to stay the same.

Adds a `ValueArray`, which encapsulates common access patterns for parsed record nodes.  `ValueArray`'s methods return instances of itself, as they intended to be chained. E.g. 

```ruby 
val.field(:creator).select { |v| v.some_attribute == 'attribute value' }
```

`ValueArray` should be extended over time to wrap more of `Array`'s functionality and capture any emerging patterns.

Lastly, this enhances `ChildDeclaration` to accept optional `:each` and `:as` options.  These options specify that an instance of the requested child class will be created for each value in the `Proc` (or `RecordProxy`) passed to `:each`, and that the matching values will be accessible as a method registered in the block's scope with the name passed to `:as`.  

```ruby
Krikri::Mapper.define(:esdn_mods) do
  sourceResource :class => DPLA::MAP::SourceResource do
    contributor :class => DPLA::MAP::Agent, :each => record.field('mods:name').select { |name| name['mods:role'].map(&:value).include?('contributor') }, :as => :contrib do
      providedLabel contrib.field('mods:namePart')
    end
  
    creator :class => DPLA::MAP::Agent, :each => record.field('mods:name').select { |name| name['mods:role'].map(&:value).include?('creator') }, :as => :creator_role do
      providedLabel creator_role.field('mods:namePart')
    end
 end
```

### DIFF
--- a/lib/krikri/mapping_dsl/parser_methods.rb
+++ b/lib/krikri/mapping_dsl/parser_methods.rb
@@ -4,9 +4,117 @@ module Krikri::MappingDSL
   module ParserMethods
     extend ActiveSupport::Concern
 
+    ##
+    # Gives access to delayed method calls on parsed record ValueArrays, to
+    # be executed at the time the mapping is processed.
+    #
+    # @return [RecordProxy] a RecordProxy providing delayed method calls against
+    #   a parsed record.
+    # @see Krikri::MappingDSL::ParserMethods::RecordProxy
     def record
-      lambda do |rec|
-        yield rec.root if block_given?
+      RecordProxy.new
+    end
+
+    ##
+    # Gives access to a delayed call for the `#local_name` of an OriginalRecord,
+    # to be executed at the time the mapping is processed.
+    #
+    # @return [Proc] a proc that, when called, returns the #local_name of the
+    #   OriginalRecord associated with the parsed record passed as its argument.
+    def local_name
+      lambda do |parsed|
+        parsed.record.local_name
+      end
+    end
+
+    ##
+    # Gives access to a delayed call for the `#rdf_subject` of an
+    # OriginalRecord, to be executed at the time the mapping is processed.
+    #
+    # @return [Proc] a proc that, when called, returns the #rdf_subject of the
+    #   OriginalRecord associated with the parsed record passed as its argument.
+    def record_uri
+      lambda do |parsed|
+        parsed.record.rdf_subject
+      end
+    end
+
+    ##
+    # This class acts as a proxy for a parsed record's nodes, wrapped in the
+    # class passed as the second argument. All methods available on the wrapper
+    # class are accepted via #method_missing, added to the #call_chain, and
+    # return `self`, allowing chained method calls.
+    #
+    #   record.field('dct:title').field('foaf:name')
+    #
+    # @see Krikri::Parser::ValueArray
+    class RecordProxy
+      attr_reader :value_class, :call_chain
+
+      ##
+      # Create a new RecordProxy object.
+      #
+      # @param call_chain [Array<Hash>] an array of hashes representing method
+      #   calls. Hashes need a :name (method name), :args (array of arguments),
+      #   and :block (a Proc to pass as a block). Defaults to [].
+      # @param klass [Class] a Class that acts as the target for delayed method
+      #   calls. Must respond to #build(record) and #values. Defaults to
+      #   Krikri::Parser::ValueArray
+      #
+      # @return [RecordProxy]
+      def initialize(call_chain = [], klass = Krikri::Parser::ValueArray)
+        @call_chain = call_chain
+        @value_class = klass
+      end
+
+      def dup
+        RecordProxy.new(call_chain.dup, value_class)
+      end
+
+      ##
+      # Wraps a given record in #value_class and applies the call chain;
+      # each method is sent to the result of the previous method. Finally,
+      # calls #values on the result.
+      #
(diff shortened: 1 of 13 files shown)


lib/krikri/parser.rb
```diff
@@ -103,5 +103,40 @@ def get_child_nodes(name)
         raise NotImplementedError, "Can't access property #{name}"
       end
     end
+
+    class ValueArray < Array
```
anarchivist: This class should have more documentation in the code and for its methods.

lib/krikri/mapping_dsl/parser_methods.rb
```diff
@@ -5,8 +5,54 @@ module ParserMethods
...
+      end
+    end
+
+    def record_uri
+      lambda do |parsed|
+        parsed.record.rdf_subject
+      end
+    end
+
+    class RecordProxy
```
anarchivist: This class should have more documentation in the code and for its methods.

anarchivist:
Looks good; as I mentioned, I think we need more documentation in the code. :book: The good news is that I think we can reuse a lot of what you used in your description of the pull request (at least at the class level).

lib/krikri/parser.rb
```diff
@@ -103,5 +103,40 @@ def get_child_nodes(name)
...
+          result = result.get_field(name)
+        end
+        result
+      end
+
+      def select(*)
+        self.class.new(super)
+      end
+
+      def match_attribute(name, other)
```
anarchivist: After looking at the [gist of the ESDN mapping](https://gist.github.com/no-reply/a3369d930c7cba223138), I realized I wanted to ask whether `Krikri::Parser::ValueArray#match_attribute` calls can be chained together, e.g., ESDN's `isShownAt` should actually be more like this based on the crosswalk:

```ruby
  isShownAt :class => DPLA::MAP::WebResource do
    uri record.field('mods:location', 'mods:url').match_attribute(:usage, 'primary').match_attribute(:access, 'object in context')
  end
```

anarchivist:
For anyone doing code review, @no-reply's [gist of the updated ESDN mapping](https://gist.github.com/no-reply/a3369d930c7cba223138) using these new DSL methods is a good reference for how these methods would be used.

lib/krikri/parser.rb
no-reply: That should absolutely work.  They should also be chainable with field selectors as in `uri record.field('mods:location', 'mods:url').match_attribute(:usage, 'primary').field('mods:someField).match_attribute(:access, 'object in context')`

anarchivist:
Other than docs, I'm :+1: if someone else takes a gander - maybe @markbreedlove, once he feels better?

no-reply:
Let me know if things look suitably documented now.

anarchivist:
Documentation looks good. I'd like a second person to review, but I'm feeling like this is good to go.

markbreedlove:
Hi:  This is proving difficult for me to digest at the moment, with limited time, though I have made some progress understanding the mappings outside of this changeset.  My main issue with this PR and the mapping in general is it's not clear as an outsider what inroads to take to get more familiar with it, due to its complexity.   I feel better now with it, having gone back and forth over a number of files to trace how the mappings are defined (e.g. to see how method_missing is used in mapping_dsl.rb to set properties) and how records are processed, but I'll have to come back to this later to get a better understanding.  I'm afraid I can't really say what I think about how this changes the mapping, yet.

lib/krikri/parser.rb
```diff
@@ -103,5 +103,75 @@ def get_child_nodes(name)
...
+    # A specialized Array object for containing Parser::Values. Provides methods
+    # for accessing and filtering values that can be chained.
+    #
+    #     my_value_array.field('dc:creator', 'foaf:name')
+    #       .match_attribute('first_name').values
+    #
+    # Methods defined on this class should return another ValueArray, an Array
+    # of literal values (retrieved from Parser::Value#value), or a single
+    # literal value.
+    class ValueArray < Array
```
anarchivist: As discussed in [Slack](https://dpla.slack.com/archives/tech/p1422394042000201), `Krikri::Parser::ValueArray` should probably not subclass `Array`, based on undesirable side effects. I'll take an attempt at another pass at this using a delegator pattern. 

We should fix this before merging; I'll give it a shot by this evening.

anarchivist:
@markbreedlove OK - thanks for your feedback. This can be difficult to review. We can talk more (with @no-reply) if you'd like. The mapping code may make more sense once we get to the point where you have a chance to contribute to the mappings.

anarchivist:
OK, as noted in Slack, I'm good on this since we're no longer subclassing `Array`. I think this is probably ready to :squash:. Ideally, I'd like to get this merged soon, but we can talk about how much more review we want this to go through.

anarchivist:
As mentioned on today's checkin call, I'm good on this being merged now, once we figure out how much we want to squash.

no-reply:
I think that's good.  I cleaned up the commit messages a bit, too.

Thanks for your help on this. It has improved appreciably while under review.

PR #62, opened 2015-01-21
Add strip whitespace and punctuation enrichments

no-reply:
This closes https://www.pivotaltracker.com/story/show/83402006


anarchivist:
Looks good to me. Just to confirm are we sure want to strip punctuation from the middle of strings too? (I'm not sure that we were doing that before, which is why I ask.)

no-reply:
I checked in with content about it seems like a :+1:.  We can always have a second enrichment that just does `#strip` and/or that does the above but leaves internal newlines.

no-reply:
:squash:ed

PR #63, opened 2015-01-21
Improve install generator and tweak views

anarchivist:


app/views/krikri/harvest_sources/index.html.erb
```diff
@@ -21,5 +26,6 @@
 
 <%= link_to 'New',
     new_institution_harvest_source_path(institution_id: @institution.id)
-%>
+%> |
```
no-reply: Is this pipe just to separate links?  Would this be better in a stylesheet?

anarchivist: Yes. And probably. :) I can consult with @AudreyAltman and push up a fix. (Alternately I can break out this out into a separate PR.)

AudreyAltman: I think we had decided to wait on all non-urgent styling. We can fix it now if you would like, but I would be okay with leaving it until we start styling work in earnest.

anarchivist: @AudreyAltman :+1: OK, no sweat. I'll close this and resubmit the PR with just one commit for the generator.

no-reply: :cool: :sunglasses:

PR #64, opened 2015-01-21
Add enrichment to strip HTML tags

no-reply:
Closes https://www.pivotaltracker.com/story/show/85082098

### DIFF
--- a/spec/lib/krikri/enrichments/strip_html_spec.rb
+++ b/spec/lib/krikri/enrichments/strip_html_spec.rb
@@ -1,18 +1,21 @@
+# -*- coding: utf-8 -*-
 require 'spec_helper'
 
 describe Krikri::Enrichments::StripHtml do
   it_behaves_like 'a field enrichment'
 
-  let(:start_value) { '<html>Moomin <i><b>Valley</i></b>' }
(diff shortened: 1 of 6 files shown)


spec/lib/krikri/enrichments/strip_html_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'spec_helper'
+
+describe Krikri::Enrichments::StripHtml do
+  it_behaves_like 'a field enrichment'
+
+  let(:start_value) { '<html>Moomin <i><b>Valley</i></b>' }
```
no-reply: You have me worried about unicode, now. Checking in on it here.

anarchivist:
I'm :+1: if there are no changes or concerns about Unicode.

no-reply:
The latest commit confirms unicode is okay, and also fixes up test duplication.

I'd prefer to keep the two commits separate, barring objections.

anarchivist:
No objections from me. @markbreedlove and @AudreyAltman?

AudreyAltman:
:+1: I like what you've done with the shared examples specs.

anarchivist:
:hospital:

PR #67, opened 2015-01-23
CouchDB Harvester

anarchivist:
This is now ready for review. Thanks to @no-reply for help on figuring out what was going on with the specs.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,12 +6,13 @@ gemspec
 gem 'engine_cart'
 
 gem 'rubocop', require: false
-
+gem 'yajl-ruby', require: 'yajl'
 gem 'database_cleaner', '~> 1.3.0', require: false
 
 group :development do
   gem 'guard'
   gem 'guard-rspec', require: false
+  gem 'pry-rescue'
 end
 
 group :test do
--- a/krikri.gemspec
+++ b/krikri.gemspec
@@ -27,6 +27,8 @@ Gem::Specification.new do |s|
   s.add_dependency "jsonpath"
   s.add_dependency "devise", "~>3.4.1"
   s.add_dependency "resque", "~>1.0"
+  s.add_dependency "dpla-couchrest", "1.2.1.pre.dpla"
+  s.add_dependency "yajl-ruby"
 
   s.add_development_dependency "sqlite3"
   s.add_development_dependency "jettywrapper", '~> 2.0'
--- a/krikri.gemspec
+++ b/krikri.gemspec
@@ -27,7 +27,7 @@ Gem::Specification.new do |s|
   s.add_dependency "jsonpath"
   s.add_dependency "devise", "~>3.4.1"
   s.add_dependency "resque", "~>1.0"
-  s.add_dependency "dpla-couchrest", "1.2.1.pre.dpla"
+  s.add_dependency "dpla-analysand", "3.1.0.pre.dpla.1"
   s.add_dependency "yajl-ruby"
 
   s.add_development_dependency "sqlite3"
--- a/lib/krikri/harvesters/couchdb_harvester.rb
+++ b/lib/krikri/harvesters/couchdb_harvester.rb
@@ -1,0 +1,115 @@
+require 'couchrest'
+require 'mime/types'
+
+module CouchRest
+  class Streamer
+    ##
+    # CouchRest::Streamer relies on curl. Some CouchDB servers behind HTTPS
+    # may have self-signed certificates, so we're adding the `--insecure`
+    # parameter to ignore strict certificate checking. There is probably a
+    # more elegant way to do this. If we *do* need to monkey patch this, it
+    # probably should be done in Heidrun and not Krikri.
+    #
+    # Note that we'll probably have to do something similar for #get_record
+    # requests: https://github.com/rest-client/rest-client/issues/288
+    def initialize
+      self.default_curl_opts = [
+        '--silent',
+        '--insecure',
+        '--no-buffer',
+        '--tcp-nodelay',
+        '-H "Content-Type: application/json"'
+      ].join(' ')
+    end
+  end
+end
+
+module Krikri::Harvesters
+  ##
+  # A harvester implementation for CouchDB
+  class CouchdbHarvester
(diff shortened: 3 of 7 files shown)


Gemfile
```diff
@@ -12,6 +12,7 @@ gem 'database_cleaner', '~> 1.3.0', require: false
 group :development do
   gem 'guard'
   gem 'guard-rspec', require: false
+  gem 'pry-rescue'
```
no-reply: Do you want to leave this in?  I'm okay with it, just want to check that you intend it to be present.

anarchivist: I didn't intend to leave it there. I can remove it if people prefer, but I've found `pry-rescue` useful in debugging.

krikri.gemspec
```diff
@@ -27,6 +27,7 @@ Gem::Specification.new do |s|
   s.add_dependency "jsonpath"
   s.add_dependency "devise", "~>3.4.1"
   s.add_dependency "resque", "~>1.0"
+  s.add_dependency "dpla-couchrest", "1.2.1.pre.dpla"
```
no-reply: What's the official policy/strategy for `dpla-couchrest`?  Do we want to cut a (non-`pre`) release and/or version it in some way that we can update, in case we need to make changes?

anarchivist: I pushed up a version of `dpla-couchrest` to [Rubygems](https://rubygems.org/gems/dpla-couchrest), and we have a [repo](https://github.com/dpla/couchrest). The only change that I made was to update the dependencies to avoid the `mime-types` version conflict that someone tried to address with couchrest/couchrest#103.

no-reply: ah. that's a simpler change than I imagined.  I'm okay with this as is, then.

anarchivist: Here are the specific changes, for the curious: https://github.com/dpla/couchrest/compare/dpla-fork

lib/krikri/harvesters/couchdb_harvester.rb
```diff
@@ -0,0 +1,115 @@
...
+      @client = CouchRest.database(uri)
+    end
+
+    ##
+    # Streams a response from a CouchDB view to yield identifiers.
+    #
+    # The following will only send requests to the endpoint until it
+    # has 1000 record ids:
+    #
+    #     record_ids.take(1000)
```
no-reply: This is an oversight of mine, but I don't think we have an actual test for this.  Is this the case, here?  Do we want `Enumerator::Lazy.new` below?

anarchivist: The [spec for `#record_ids` in the harvester shared examples](https://github.com/dpla/KriKri/blob/feature/couchdb-harvester/spec/support/shared_examples/harvester.rb#L21-L29) checks for this to be an `Enumerator`.

no-reply: We should probably update it to ask for an `Enumerator::Lazy`.  I think I was trying to keep the general Harvester requirements loose, but I think it's important to have access to the lazy version of `#map`, at least.

We probably need to accept that the degree of laziness will be different from harvester to harvester, and I think that this works, as is.

anarchivist: Can you help me understand the constructor's expectation for `Enumerator::Lazy`? The documentation is pretty horrible: http://ruby-doc.org/core-2.1.3/Enumerator/Lazy.html

no-reply: Probably, but I think `#lazy` is fine, here.  The streamer ends up downloading the whole view, anyway, if I understand correctly.

anarchivist: OK - I was a little confused by your feedback. Should I work on instantiating `Enumerator::Lazy` or not?

no-reply: Not.  I just mean in the spec: https://github.com/dpla/KriKri/blob/feature/couchdb-harvester/spec/support/shared_examples/harvester.rb#L21-L29

anarchivist: WOOHOO!

anarchivist:
After thinking about this this evening, it might be possible to reimplement this using a slightly simpler set of requests using `Yajl::HttpStream`.

no-reply:
Is this squashed because it's ready to go? Or do you want to play with `HttpStream` more?

I'm :+1: on this, it seems likely that it will get us through our first harvest (unless you think your parse issues are more than a connection problem).

anarchivist:
I took a stab at refactoring this using a different client library ([Analysand](https://github.com/yipdw/analysand]), which can be seen in 4c12f430703a154516131b31e984b11c8fe44cfc (it's in a separate branch, `features/couchdb-harvester-refactor`). I don't know if one will prove to be more robust than the other, but the Analysand-based implementation feels a little cleaner.

no-reply:
That looks better for me, at least on this side.

anarchivist:
I would propose that I push this commit to the branch associated with this PR, and that we not squash the commits (in the case that we need to refer to the initial implementation.) If you're OK with that, then I think that this is close to ready to merge. 

I'll wait for the green light for you before I push to this branch.

no-reply:
:+1:

anarchivist:
Ready to :merge:.

no-reply:
Looking underneath the hood on `Analysand`, I'm an even bigger :+1: on this than before.

Anyone want to give it a quick look?  If not, I'm good to :merge:.
:rocket:

Gemfile
```diff
@@ -6,12 +6,13 @@ gemspec
 gem 'engine_cart'
 
 gem 'rubocop', require: false
-
+gem 'yajl-ruby', require: 'yajl'
```
no-reply: @anarchivist I know this is already merged, but can you say something about this?  Will the correct thing get required when this is installed by a host app?

anarchivist: Yes. The gem is included in the gemspec as well, but it gets required using `require 'yajl'`, not `require 'yajl-ruby'`.

PR #68, opened 2015-01-26
Add Krikri::Parsers::QdcParser

anarchivist:
Adds a qualified Dublin Core parser. The default root node path is the
one used by CONTENTdm. This can be overridden (like all Krikri::XmlParsers)
at the time of instantiation.

### DIFF
--- a/lib/krikri.rb
+++ b/lib/krikri.rb
@@ -12,4 +12,5 @@ module Krikri
   autoload :OaiDcParser,    'krikri/parsers/oai_dc_parser'
   autoload :JsonParser,     'krikri/parsers/json_parser'
   autoload :ModsParser,     'krikri/parsers/mods_parser'
+  autoload :QdcParser,      'krikri/parsers/qdc_parser'
 end
--- a/lib/krikri/parsers/qdc_parser.rb
+++ b/lib/krikri/parsers/qdc_parser.rb
@@ -1,0 +1,18 @@
+module Krikri
+  ##
+  # A Qualified Dublin Core parser. Uses XML parser with a root path to match
+  # the metadata path as harvested. The default root_path is the element used
+  # by CONTENTdm; this can be overridden as with all Krikri::XmlParsers, at
+  # the time of instantiation.
+  #
+  # @see Krikri::XmlParser
+  class QdcParser < XmlParser
+    def initialize(record, root_path = '//oai_qdc:qualifieddc', ns = {})
+      ns = {
+        qdc: 'http://epubs.cclrc.ac.uk/xmlns/qdc/',
+        oai_qdc: 'http://worldcat.org/xmlschemas/qdc-1.0/'
+      }.merge(ns)
+      super(record, root_path, ns)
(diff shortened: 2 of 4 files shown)


no-reply:
Looks good to me. :rocket:

lib/krikri/parsers/qdc_parser.rb
```diff
@@ -0,0 +1,18 @@
+module Krikri
+  ##
+  # A Qualified Dublin Score parser. Uses XML parser with a root path to match
```
AudreyAltman: Typo in line above: should be `Qualified Dublin Core`?

anarchivist: eep :weary:

lib/krikri.rb
```diff
@@ -12,4 +12,5 @@ module Krikri
   autoload :OaiDcParser,    'krikri/parsers/oai_dc_parser'
   autoload :JsonParser,     'krikri/parsers/json_parser'
   autoload :ModsParser,     'krikri/parsers/mods_parser'
+  autoload :QdcParser,      'krikri/parsers/qdc_parser'
 end
```
AudreyAltman: I am confused about autoloading.  I thought that this line from `lib/krikri/engine.rb` autoloads everything in `lib` and its subdirectories recursively, including the parsers:
`config.autoload_paths += Dir["#{config.root}/lib/**/"]`
Am I incorrect?

no-reply: I believe you have it right.  That line went in with little fanfare in a PR that was parallel to some of the parser work, and it has been frequently overlooked by just about everyone since.  I think we should be able remove all of these autoload lines.

AudreyAltman:
Aside from the `autoload` question, this code looks ship-shape to me.

lib/krikri.rb
anarchivist: As noted in Slack, removing these autoload declarations doesn't lead to the parsers to get loaded.

no-reply: My vote would be to punt it. Open a chore to figure it out in another PR.

anarchivist: I took a look at our use of `autoload` in other modules; currently, only [`krikri/ldp`](https://github.com/dpla/KriKri/blob/develop/lib/krikri/ldp.rb) and [`krikri/enrichments`](https://github.com/dpla/KriKri/blob/develop/lib/krikri/enrichments.rb) are using them. We don't have a `Krikri::Parsers` module, however.

no-reply: Oh... the line is `config.autoload_paths += Dir["#{config.root}/lib/**/"]` not `.../**/*`.

It's not including subdirs. LDP and Enrichments are requiring everything below them, but Parsers is not.

anarchivist: See [#86959230](https://www.pivotaltracker.com/story/show/86959230)

anarchivist:
Just :squash:'ed to fix the typo.

PR #69, opened 2015-01-27
Add convenience constants to Krikri::MappingDSL

anarchivist:
* Add Krikri::MappingDSL::DCMITYPE_LABELS
* Add Krikri::MappingDSL::GENRE_LABELS

Note that the proposed genre labels may be out of alignment with MAPv4, and `dpla/map` (which suggests that all terms assigned to `genre`/`edm:hasType` should be from AAT).

### DIFF
--- a/lib/krikri/mapping_dsl.rb
+++ b/lib/krikri/mapping_dsl.rb
@@ -13,6 +13,14 @@ module Krikri
     include ParserMethods
     include RdfSubjects
 
+    DCMITYPE_LABELS = RDF::DCMITYPE.select(&:class?).map { |t| t.label.downcase }
+
+    # TODO: Discuss with Content Team. `dpla_map` currently constrains
+    # genre (edm:hasType) to AAT terms; these are not all AAT terms.
+    GENRE_LABELS = ['book', 'film/video', 'manuscript', 'maps', 'music',
+                    'musical score', 'newspapers', 'nonmusic',
+                    'photograph/pictorial works', 'serial']
+
     def properties
       @properties ||= []
     end


lib/krikri/mapping_dsl.rb
```diff
@@ -13,6 +13,14 @@ module MappingDSL
     include ParserMethods
     include RdfSubjects
 
+    DCMITYPE_LABELS = RDF::DCMITYPE.select(&:class?).map { |t| t.label.downcase }
```
no-reply: :cool:

lib/krikri/mapping_dsl.rb
```diff
@@ -13,6 +13,14 @@ module MappingDSL
...
+
+    # TODO: Discuss with Content Team. `dpla_map` currently constrains
+    # genre (edm:hasType) to AAT terms; these are not all AAT terms.
+    GENRE_LABELS = ['book', 'film/video', 'manuscript', 'maps', 'music',
```
no-reply: I think it may be worth discussing this with Gretchen and @rudeamy. I'm not going to argue for AAT only, here; the status of http://github.com/dpla/dpla_map is really down to the limitations of its validation capabilities.  However, this does raise the question (again) of whether this kind of filter/validation/enforcement of a concept scheme belongs in mapping or enrichment.

I'm tempted to say that these allowed value lists *should* be accurately reflected in `dpla-map`, values cleaned up in the enrichment process.

anarchivist:
Since we decided this was an enrichment, I'm going to close this PR.

PR #70, opened 2015-01-27
Feature/date normalization

no-reply:
Attempts to convert a string value to a Date object:
   - Parses EDTF values, returns an appropriate EDTF object if a match is found; otherwise...
   - Parses values in %m*%d*%Y format and returns a Date object if appropriate; otherwise...
   - Parses values that match any of Date#parse's supported formats.

If the value is not a `String` or is parsed as invalid by all parsers, the original value is returned unaltered.

### DIFF
--- a/krikri.gemspec
+++ b/krikri.gemspec
@@ -23,6 +23,7 @@ Gem::Specification.new do |s|
   s.add_dependency "rdf-marmotta", '>= 0.0.6'
   s.add_dependency "blacklight", "~>5.8.0"
   s.add_dependency "therubyracer"
+  s.add_dependency "edtf"
   s.add_dependency "oai"
   s.add_dependency "jsonpath"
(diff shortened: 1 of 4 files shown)


anarchivist:
:+1:.

Did you look at [Timeliness](https://github.com/adzap/timeliness) at all for handling the parsing? (I know that's what BPL is using in [`bpl_enrich`](https://github.com/boston-library/bpl_enrich), and I don't know if we should be that concerned about speed, etc. up front.)

no-reply:
Most of the stuff in `Timeliness` is already handled by `Date.parse` (at least as far as Dates go).  

See EDTF aware work, here:

https://github.com/dpla/KriKri/blob/ab06c194c26516f4f601d82a1c055fc86abd587f/lib/krikri/util/extended_date_parser.rb

PR #71, opened 2015-01-29
Date Enrichments

no-reply:
Implements two date enrichments:

  - `ParseDate` converts string values in individual fields to `Date` objects.  This enrichment is meant to be run on TimeSpan attributes `#begin`/`#end` when they have been mapped directly.
  - `TimespanSplit` accepts a string value or a TimeSpan, and returns a timespan object with a single `#begin` and `#end`, if any can be found.  If no `#begin` and `#end` are set, they are parsed from `#providedLabel`.

Both rely on `Krikri::Util::ExtendedDateParser`, which parses strings to one of `Date`, `EDTF::Interval`, or `EDTF::Epoch`.  The last two represent date ranges; when they are returned to `TimespanSplit` it uses their methods to determine the start and end of the range.

### DIFF
--- a/krikri.gemspec
+++ b/krikri.gemspec
@@ -23,6 +23,7 @@ Gem::Specification.new do |s|
   s.add_dependency "rdf-marmotta", '>= 0.0.6'
   s.add_dependency "blacklight", "~>5.8.0"
   s.add_dependency "therubyracer"
+  s.add_dependency "edtf"
   s.add_dependency "oai"
   s.add_dependency "jsonpath"
   s.add_dependency "devise", "~>3.4.1"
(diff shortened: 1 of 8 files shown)


no-reply:
This supercedes and closes #70.

no-reply:
Note that though this uses EDTF internally, all data is eventually converted to `xsd:Date` when serialized to RDF.  We need to do some work to implement an EDTF `RDF::Literal` to utilize EDTF more generally.

anarchivist:
:+1: I think this is ready to squash; took me a little while to understand how to call these.

no-reply:
:squash:'d

anarchivist:
:clock1: :clock1030: :money_with_wings:

PR #72, opened 2015-01-30
Add a genre filter enrichment

no-reply:
Removes values that do not match a given list of genre terms.

### DIFF
--- a/lib/krikri/enrichments/genre_filter.rb
+++ b/lib/krikri/enrichments/genre_filter.rb
@@ -1,0 +1,46 @@
+module Krikri::Enrichments
+  ##
+  # Enrichment to remove non-genre fields from
+  #
+  #   StripHtml.new.enrich_value('Book') => 'Book'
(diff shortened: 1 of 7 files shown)


lib/krikri/enrichments/genre_filter.rb
```diff
@@ -0,0 +1,46 @@
...
+  #  - Maps
+  #  - Music
+  #  - Musical Score
+  #  - Newspapers
+  #  - Nonmusic
+  #  - Photograph/Pictorial Works
+  #  - Serial
+  #
+  # Removes all non-string values
+  class GenreFilter
```
anarchivist: I might call this `DPLAGenreFilter`, just cuz.

no-reply: I think we should actually move the whole thing to `Heidrun`.  It's clearly not a core krikri enrichment.

PR #78, opened 2015-02-02
Simplify CouchdbHarvester#get_record

anarchivist:
The prerelease of Analysand fixes a bug I reported where document IDs were
not getting escaped properly (yipdw/analysand#7). We don't need to call
a view request and can instead use `Analysand:Database#get!`.

Also fixes one of the CouchdbHarvester specs for `#get_record`, which was
not implemented properly.


no-reply:
This looks good to me.  :rocket:

PR #79, opened 2015-02-02
Install Jetty config using generator; tweak MARMOTTA_HOME

anarchivist:
* Install Jetty configuration when running generator
* MARMOTTA_HOME:
    * Use `#{app_root}/jetty/marmotta` by default instead of `#{app_root)/marmotta`
    * Set MARMOTTA_HOME when running rake ci


no-reply:
Awesome.

PR #80, opened 2015-02-03
Skip unavailiable fields in FieldEnrichment

no-reply:
Given a field chain like `{ :sourceResource => { :creator => :label } }`, values of `creator` that don't have a property `label` would previously cause the enrichment to fail.  Now they are simply passed over.

### DIFF
--- a/lib/krikri/field_enrichment.rb
+++ b/lib/krikri/field_enrichment.rb
@@ -39,6 +39,7 @@ module Krikri
 
     def enrich_field(record, field_chain)
       field = field_chain.first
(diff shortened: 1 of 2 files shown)


no-reply:
This prepares better for enrichments on invalid MAP records.  It's not a major fix, in that it's hard to say when we would have run into this issue in real life, but the ISO enrichment runs into it through the shared examples for field enrichments.

https://github.com/dpla/KriKri/tree/feature/iso-enrich

anarchivist:
:100:

PR #81, opened 2015-02-03
ISO Enrichment

no-reply:
Converts text fields and/or providedLabels to ISO 689-3 URIs (Lexvo), searching for exact matches on 3 letter codes and labels.

Transforms text values matching either (English) labels or ISO language codes from Lexvo into DPLA::MAP::Controlled::Language resources with appropriate URIs.

For example:
```ruby
      'fin'     => 'http://lexvo.org/id/iso639-3/fin'
      'finnish' => 'http://lexvo.org/id/iso639-3/fin'
      'eng'     => 'http://lexvo.org/id/iso639-3/eng'
      'english' => 'http://lexvo.org/id/iso639-3/eng'
```
If no matches are found, returns a bnode with the input value as providedLabel.

If passed an ActiveTriples::Resource, the enrichment will:
  - Perform the above text matching on any present `providedLabel`s,  returning the original node if no results are found.
    - If a match is found, the resource is replaced by one with the appropriate URI.
    - If multiple values are provided and multiple matches found, they will be deduplicated.
  - Leave DPLA::MAP::Controlled::Language objects that are not bnodes unaltered.
  - Remove any values which are not either bnodes or members of  DPLA::MAP::Controlled::Language.

I'll discuss it with @guegueng and @rudeamy on 2/3.

### DIFF
--- a/lib/krikri/enrichments/iso_enrich.rb
+++ b/lib/krikri/enrichments/iso_enrich.rb
@@ -17,7 +17,7 @@ module Krikri::Enrichments
   #
   # If passed an ActiveTriples::Resource, the enrichment will:
   #
-  #   - Perform the above text matching on any present `provededLabel`s,
+  #   - Perform the above text matching on any present `providedLabel`s,
   #     returning the original node if no results are found.  If multiple
   #     values are provided and multiple matches found, they will be
   #     deduplicated.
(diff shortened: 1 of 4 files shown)


no-reply:
Note that this *does not* currently add a `prefLabel` to exact language matches.

no-reply:
After discussion w/ @guegueng and @rudeamy, the verdict is to rewrite so `exactMatches` use the form: _:agg1 dc:language _:lang_node1 .
_:lang_node1 skos:exactMatch lexvo:eng .
_:lang_node1 dpla:providedLabel 'eng' .
lexvo:eng skos:prefLabel 'english'  .

no-reply:
Closing per https://github.com/dpla/KriKri/pull/81#issuecomment-72689998. Will resubmit with changes to dpla/heidrun.

PR #82, opened 2015-02-03
Change dependency from dpla-analysand to analysand

anarchivist:
Analysand 4.0.0 (officially released) fixes the workaround which introduced the
dependency on our prerelease for 1268f3b2a509b2cba0364d243722680733b9b0a7.


no-reply:
Everything checks out here. :white_check_mark:

PR #84, opened 2015-02-04
Add error logging to Harvester

no-reply:
~~I'm submitting this mostly for input.  @markbreedlove: thoughts?~~

Now with more complete error handling; more or less as discussed at https://dpla.slack.com/archives/tech/p1423166676000192.

 - Catches errors on individual records during `Harvester` runs, logging and passing over failed records.

 - Adds logging to `Activity#run`. Failed jobs are logged, the error is rethrown and the `end_time` is set for both successes and failures. Sets `end_time` to `nil` when a "ended" job restarts, so `end_time` should never be earlier than `start_time`.

### DIFF
--- a/lib/krikri/harvester.rb
+++ b/lib/krikri/harvester.rb
@@ -102,7 +102,14 @@ module Krikri
     # @return [Boolean]
     def run(activity_uri = nil)
       log :info, 'harvest is running'
-      records.each { |rec| rec.save(activity_uri) }
+      records.each do |rec|
+        begin
+          rec.save(activity_uri)
+        rescue => e
+          log :error, "Error harvesting record:\n#{rec.content}\n\twith message:\n#{e.message}"
+          next
+        end
+      end
       log :info, 'harvest is done'
       true
     end
--- a/app/models/krikri/activity.rb
+++ b/app/models/krikri/activity.rb
@@ -42,10 +42,28 @@ module Krikri
     # is passed an instance of the agent, and a URI representing this Activity.
     def run
       if block_given?
+        update_attribute(:end_time, nil) if ended?
         set_start_time
-        yield agent_instance, rdf_subject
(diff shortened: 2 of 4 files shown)


lib/krikri/harvester.rb
```diff
@@ -102,7 +102,12 @@ def get_record(_)
     # @return [Boolean]
     def run(activity_uri = nil)
       log :info, 'harvest is running'
-      records.each { |rec| rec.save(activity_uri) }
+      begin
```
no-reply: or maybe we want:

```ruby
records.each do |rec|
  begin
    rec.save(activity_uri)
  rescue => e
    log :error, "Error harvesting record: \n #{record.content} \n\twith message #{e.message}"
  end
end```

anarchivist: I think that sounds reasonable. I'd love to talk through more how errors should get handled by harvesters and when we want to fail the whole thing.

markbreedlove: It's nice to have a string that we can reliably grep on, so your example above would be an improvement.

no-reply:
This is now ready for a more serious review.  See updated PR description.

markbreedlove:
I'm happy with that.  I like that the activity is resetting its end timestamp if it's being re-run.

I'm also OK with letting individual records fail in order to gather more information.  When we build out the events, we can have them stop and not continue with the next stage (e.g. enrichment) if there were errors, or maybe if there was a threshold exceeded.  The U.I. is the place for prompting about what to do about errors.

no-reply:
:rocket: ?

anarchivist:
:+1: :exclamation: :squash: :question:

no-reply:
:merge:

anarchivist:
:tractor: :runner: :memo:

PR #85, opened 2015-02-04
Retry LDP requests on certain Errors

no-reply:
This will retry the errors we have been seeing a maximum of four times, starting with a 0.025-0.0375 second delay, and doubling that interval for each retry.  The longest delay would be 0.2-0.3 seconds.


anarchivist:
Do we need to modify the specs for this?

no-reply:
Should probably mock some of the errors, yes.

markbreedlove:
That sounds good to me.

no-reply:
This is ready for review.

The non-`ConnectionError` retry classes are the defaults.

anarchivist:
:+1: :squash: :question:

no-reply:
:white_check_mark: :white_flower:

anarchivist:
:recycle:

PR #86, opened 2015-02-04
Compare records

AudreyAltman:
View original record and enhanced record side-by-side
* SearchIndexDocument model (subclass of SolrDocument) allows a single document returned from a Solr query to be associated with its DPLA::MAP::Aggregation
* New initializer mix-in to DPLA::MAP::Aggregation allows aggregation to be associated with its original record
* Allows OriginalRecord to be loaded from an identifier with a mime type extension
* Rake tasks save sample aggregations and original records to Marmotta
* New records#show view with helper methods in SearchResultsHelperBehavior
* IndexService does not attempt to index fields that are not defined in the Solr schema

To test in browser:
* Generate test app with engine cart
* You may need to `stop`, `clean`, and `config` jetty
* Follow new instructions in `README.md` to save a sample valid and/or invalid record
* Open `/krikri/records` in your browser and click on a record to view `records#show`

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,22 +95,25 @@ models, and views.  For more information, see [Blacklight's website](https://git
 Sample Data
 -----------
 
-To index a sample aggregation into solr, from `/krikri/spec/internal`:
-    rake krikri:index_sample_aggregation
+To save a sample record to Solr and Marmotta, from `/krikri/spec/internal`:
 
-To index an _invalid_ sample aggregation into solr:
-    rake krikri:index_invalid_aggregation
+    rake krikri:samples:save_record
 
-To delete the sample aggregation:
-    rake krikri:delete_sample_aggregation
+To save an _invalid_ sample record to Solr and Marmotta:
 
-To create a sample institution and harvest source, from `/krikri/spec/internal`:
+    rake krikri:samples:save_invalid_record
 
-    rake krikri:create_sample_institution
+To delete all sample records:
+
+    rake krikri:samples:delete_record
+
+To save a sample institution and harvest source, from `/krikri/spec/internal`:
+
+    rake krikri:samples:save_institution
 
 To delete the sample institution and harvest source:
 
-    rake krikri:delete_sample_institution
+    rake krikri:samples:delete_institution
 
 
 Using Vagrant for development (experimental)
--- a/app/views/krikri/records/show.html.erb
+++ b/app/views/krikri/records/show.html.erb
@@ -1,1 +1,19 @@
-This view will show an enriched record alongside its original record.
+<% content_for :title do %>QA Record: <%= @document.id.html_safe %><% end %>
+
+<div id="main-container" class="container">
+  <%= render :partial=>'/flash_msg', layout: 'shared/flash_messages' %>
+
+  <h2>QA Record: <%= @document.id.html_safe %></h2>
+
+  <div class="row">
+    <div class="col-md-6 col-sm-6">
+      <h3>Enriched Record</h3>
+      <pre class="pre-scrollable"><%= render_enriched_record(@document) %></pre>
+    </div>
+
+    <div class="col-md-6 col-sm-6">
+      <h3>Original Record</h3>
+      <pre class="pre-scrollable"><%= render_original_record(@document) %></pre>
+    </div>
+  </div>
+</div>
--- a/lib/krikri/search_results_helper_behavior.rb
+++ b/lib/krikri/search_results_helper_behavior.rb
@@ -42,5 +42,60 @@ module Krikri
       false
     end
 
+    # Render enriched record for view
+    # @param [Krikri::SearchIndexDocument]
+    # @return [String]
+    def render_enriched_record(document)
+      agg = document.aggregation
+      return error_msg('Aggregation not found.') unless agg.present?
+      JSON.pretty_generate(agg.to_jsonld)
+    end
+
+    # Render original record for view
+    # @param [Krikri::SearchIndexDocument]
+    # @return [String]
+    def render_original_record(document)
+      agg = document.aggregation
+      return error_msg('Aggregation not found.') unless agg.present?
+
+      begin
+        original_record = agg.original_record
+      rescue StandardError => e
+        logger.error e.message
+        return error_msg(e.message)
+      end
+
+      return error_msg('Original record not found.') unless 
+        original_record.present?
+      prettify_string(original_record.to_s, original_record.content_type)
+    end
+
+    private
+
+    def prettify_string(string, mime_type)
+      string = prettify_json_string(string) if mime_type.include? 'json'
+      string = prettify_xml_string(string) if mime_type.include? 'xml'
+      string
+    end
+
+    def prettify_json_string(string)
+      begin
+        return JSON.pretty_generate(JSON.parse(string))
+      rescue JSON::ParserError
+        return string
+      end
+    end
+
+    def prettify_xml_string(string)
+      if Nokogiri.XML(string).errors.empty?
+        doc = Nokogiri.XML(string) { |c| c.noblanks }
+        return doc.to_xml(indent: 2)
+      end
+      string
+    end
+
+    def error_msg(message = '')
+      "There was a problem getting the record.\n\n#{message}"
+    end
   end
 end
--- a/lib/tasks/krikri.rake
+++ b/lib/tasks/krikri.rake
@@ -1,77 +1,115 @@
 require 'factory_girl_rails'
 include FactoryGirl::Syntax::Methods
 require 'dpla/map/factories'
+require '../../app/models/krikri/original_record.rb'
+require '../../spec/factories/krikri_original_record.rb'
 require 'open-uri'
 require 'resque/tasks'
 
 require 'krikri/index_service'
 
 namespace :krikri do
+  namespace :samples do
 
-  # Tasks will execute in spec/internal unless directory is changed
-  desc 'Index a valid sample aggregation'
-  task :index_sample_aggregation do
-    agg = build(:aggregation)
+    # Note: The content of the sample original record does not actually
+    # correspond to the content of the sample aggregation.
+    desc 'Save a sample record to Marmotta and Solr'
+    task :save_record => :environment do
+      original_record = build(:oai_dc_record)
+      original_record.save unless original_record.exists?
 
-    graph = agg.to_jsonld['@graph'][0]
+      # Make a DPLA::MAP::Aggregation with an original record instantiated as an
+      # ActiveTriples::Resource object.
+      agg = build(:aggregation,
+                  :originalRecord => ActiveTriples::Resource
+                                  .new(original_record.rdf_subject)
+            )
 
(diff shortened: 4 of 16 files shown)


lib/krikri/search_results_helper_behavior.rb
```diff
@@ -42,5 +42,51 @@ def render_bookmarks_control?
...
+    def prettify_string(string)
+      # return JSON if original record content is JSON
+      begin
+        return JSON.pretty_generate(JSON.parse(string))
+      rescue JSON::ParserError
+      end
+
+      # return XML String if original record content is XML
+      return Nokogiri::XML(string).to_xml if
+        Nokogiri::XML(string).errors.empty?
```
AudreyAltman: This does not prettify the XML as well as I'd like.  It's inconsistent with line breaks and indentations.  But I couldn't find a better way - any suggestions would be greatly appreciated!

lib/tasks/krikri.rake
```diff
@@ -8,38 +10,73 @@ require 'krikri/index_service'
...
 
-    graph = agg.to_jsonld['@graph'][0]
+    # Make a DPLA::MAP::Aggregation with an original record instantiated as an
+    # ActiveTriples::Resource object.
+    agg = build(:aggregation,
+                :originalRecord => ActiveTriples::Resource
+                                .new(original_record.rdf_subject)
+          )
+
+    agg.mint_id!('krikri_sample')
```
no-reply: I'd love to see this as a custom build strategy.  It's repeated below, and I think it would be useful in other tests.

https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#custom-strategies

Alternatively, it occurs to me that this pattern of instantiating `Aggregation`s, then calling 'mint_id!` might deserve it's own builder method (or object?). Thoughts?

README.md
```diff
@@ -95,21 +95,19 @@ models, and views.  For more information, see [Blacklight's website](https://git
...
 
-To index a sample aggregation into solr, from `/krikri/spec/internal`:
-    rake krikri:index_sample_aggregation
+To save a sample record to Solr and Marmotta, from `/krikri/spec/internal`:
+    rake krikri:save_sample_record
 
-To index an _invalid_ sample aggregation into solr:
-    rake krikri:index_invalid_aggregation
+To save an _invalid_ sample record to Solr and Marmotta:
+    rake krikri:save_invalid_sample_aggregation
```
anarchivist: This should be `rake krikri:save_invalid_sample_record`

lib/krikri/search_results_helper_behavior.rb
```diff
@@ -42,5 +42,51 @@ def render_bookmarks_control?
       false
     end
 
+    # Render enriched record for view
+    # @param [Krikri::SearchIndexDocument]
+    # @return [String]
+    def render_enriched_record(document)
+      agg = document.aggregation
+      return error_msg unless agg.present?
+      JSON.pretty_generate(agg.to_jsonld)
```
anarchivist: Thinking out loud - do we want to present more API-like data here (e.g, with a briefer/differently mapped context)?

lib/krikri/search_results_helper_behavior.rb
```diff
@@ -42,5 +42,51 @@ def render_bookmarks_control?
...
+        logger.error e.message
+        return error_msg
+      end
+
+      prettify_string(original_record.to_s)
+    end
+
+    private
+
+    def prettify_string(string)
```
anarchivist: Should (could) `#prettify_string` take a MIME type as an argument, which gets passed from `#render_original_record`?  Then we could do a series of simple checks - attempt `JSON.parse` only if the MIME Type is, say, `application/json`, etc.

lib/krikri/search_results_helper_behavior.rb
anarchivist: Maybe something like this:

```
doc = Nokogiri.XML(x)  { |c| c.noblanks }
doc.to_xml(indent: 2)
```

lib/tasks/krikri.rake
anarchivist: +1. I think you're suggesting something similar in dpla/heidrun#14.

lib/tasks/krikri.rake
```diff
@@ -8,38 +10,73 @@ require 'krikri/index_service'
 
 namespace :krikri do
 
-  # Tasks will execute in spec/internal unless directory is changed
-  desc 'Index a valid sample aggregation'
-  task :index_sample_aggregation do
-    agg = build(:aggregation)
+  # Note: The content of the sample original record does not actually
```
anarchivist: While we're in the this tasks file, maybe we should consider namespacing these all under something like `krikri:index`, or `krikri:samples`?

lib/krikri/search_results_helper_behavior.rb
```diff
@@ -42,5 +42,51 @@ def render_bookmarks_control?
...
+      # return XML String if original record content is XML
+      return Nokogiri::XML(string).to_xml if
+        Nokogiri::XML(string).errors.empty?
+
+      # return String
+      string
+    end
+
+    def error_msg
+      'There was a problem getting the record.'
```
anarchivist: Could we include the error message somewhere, perhaps a few lines down?

README.md
```diff
@@ -95,21 +95,19 @@ models, and views.  For more information, see [Blacklight's website](https://git
 Sample Data
 -----------
 
-To index a sample aggregation into solr, from `/krikri/spec/internal`:
-    rake krikri:index_sample_aggregation
+To save a sample record to Solr and Marmotta, from `/krikri/spec/internal`:
+    rake krikri:save_sample_record
```
anarchivist: Each of these indented lines should have have a blank newline in front of them.

app/views/krikri/records/show.html.erb
```diff
@@ -1 +1,20 @@
...
+
+<div id="main-container" class="container">
+  <%= render :partial=>'/flash_msg', layout: 'shared/flash_messages' %>
+
+  <h2>QA Record: <%= @document.id.html_safe %></h2>
+
+  <div class="row">
+    <div class="col-md-6 col-sm-6">
+      <h3>Original Record</h3>
+      <pre><%= render_original_record(@document) %></pre>
```
anarchivist: Could we perhaps change this `<pre>` to `<pre class="pre-scrollable">`?

app/views/krikri/records/show.html.erb
```diff
@@ -1 +1,20 @@
...
+    </div>
+
+    <div class="col-md-6 col-sm-6">
+      <h3>Enriched Record</h3>
+        <pre><%= render_enriched_record(@document) %></pre>
```
anarchivist: Could we perhaps change this `<pre>` to `<pre class="pre-scrollable">`?

anarchivist:
Great work! I think none of my feedback should require anything super complicated. I'm also raising a few questions here and there (mostly for @no-reply).

lib/krikri/search_results_helper_behavior.rb
AudreyAltman: That did the trick!  Yay!

lib/krikri/search_results_helper_behavior.rb
AudreyAltman: :+1:

README.md
AudreyAltman: Good catch.

lib/krikri/search_results_helper_behavior.rb
AudreyAltman: Follow-up question:  I implemented this using `OriginalRecord`'s `content_type` method.  How sure are we that all original records will have a mime type definition that will be accessible with this method?  Are all original records assigned a mime type?

no-reply: All `OriginalRecord`s have a ctype by virtue of https://github.com/dpla/KriKri/blob/develop/app/models/krikri/original_record.rb#L107.

The `content_type` is used by Marmotta to determine both the file extension and the interaction mode (RDFSource v. NRSource) of the posted content; so it's fairly important that an appropriate mime type is used by the harvester.

anarchivist: In the event that the harvester doesn't appropriately set the MIME type of the data, I think Marmotta should default to `application/octet-stream`.

no-reply: Is that a "we want it to" or "it will"?  The implementation of `OriginalRecord#content_type` defaults to `text/xml`.

https://github.com/dpla/KriKri/blob/develop/app/models/krikri/original_record.rb#L107

anarchivist: _Marmotta_ will. You're right about Krikri.

lib/krikri/search_results_helper_behavior.rb
no-reply: I'm :+1: on this, but think it's outside the scope of the PR.  I think we'll want a `DPLA::MAP::Aggregation#to_map_json` that takes out the LDP data and squashes out superfluous bnode ids.

We also need to look at hosting our `@context` somewhere, so it doesn't need to appear in every serialized document.

I propose to consider this point dealt with for this PR by the creation of: https://www.pivotaltracker.com/story/show/87730010

app/views/krikri/records/show.html.erb
AudreyAltman: What a fantastic idea :confetti_ball: :)

lib/krikri/search_results_helper_behavior.rb
no-reply: Sounds like we want to consider changing the behavior of `OriginalRecord` to reflect Marmotta.  Then this code can expect `application/octet-stream`, where more specific types aren't available.

lib/krikri/search_results_helper_behavior.rb
AudreyAltman: @anarchivist I'm not clear on what you mean here.

lib/tasks/krikri.rake
no-reply: :+1: on `krikri:samples`.

lib/krikri/search_results_helper_behavior.rb
anarchivist: You're catching the exception in `#render_original_record`, and that calls `#error_msg`. Can `#render_original_record` also include the error message in what gets rendered?

lib/tasks/krikri.rake
AudreyAltman: Created new chore in Pivotal Tracker for this: https://www.pivotaltracker.com/n/projects/1172184/stories/87733566

no-reply:
This looks awesome.  Minus the few tweaks still outstanding from comments, I think it's ready to go.

I *believe* it will pass CI once rebased to a point after the merge of #88.

AudreyAltman:
Do not merge yet - awaiting feedback from Gretchen.

AudreyAltman:
As per request from Gretchen, the view layout has been changed so that enriched record appears to the left of original record.

no-reply:
:rocket:

PR #89, opened 2015-02-05
Add parser arguments to mapping definitions

no-reply:
Krikri::Mapper::Agent receives only unparsed records, assuming that the
Mapping will handle parsing through `#process_record`.  The parser to
use is specified through the mapping definition with the `:parser`
option.

Previously, there was no way to pass arguments to the parser. Arguments
are not needed in most cases, but necessary for some `Krikri::XmlParser`
subclasses.  This adds `:parser_args`, allowing you to pass an array
of parser arguments when defining a mapping.

    Krikri::Mapper.define(:scdl_qdc,
		          :parser => Krikri::QdcParser,
    			  :parser_args => ['//qdc:qualifieddc']) do
      ...
    end

### DIFF
--- a/lib/krikri/mapping.rb
+++ b/lib/krikri/mapping.rb
@@ -10,14 +10,19 @@ module Krikri
   class Mapping
     include MappingDSL
 
-    attr_reader :klass, :parser
+    attr_reader :klass, :parser, :parser_args
 
     ##
     # @param klass [Class] The model class to build in the mapping process.
     # @param parser [Class] The parser class with which to process resources.
-    def initialize(klass = DPLA::MAP::Aggregation, parser = Krikri::XmlParser)
+    # @param parser_args [Array] The arguments to pass to the parser when
+    #   processing records.
+    def initialize(klass = DPLA::MAP::Aggregation,
+                   parser = Krikri::XmlParser,
+                   *parser_args)
       @klass = klass
       @parser = parser
+      @parser_args = parser_args
     end
 
     ##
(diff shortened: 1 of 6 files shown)


anarchivist:
I'm :+1: on this, but let me take a look in the morning once I'm rested. :zzz:

no-reply:
I want to add a commit message, too.

anarchivist:
:speech_balloon:

no-reply:
:eyes:

lib/krikri/mapping.rb
```diff
@@ -10,14 +10,17 @@ module Krikri
   class Mapping
     include MappingDSL
 
-    attr_reader :klass, :parser
+    attr_reader :klass, :parser, :parser_args
 
     ##
     # @param klass [Class] The model class to build in the mapping process.
     # @param parser [Class] The parser class with which to process resources.
```
anarchivist: Need to add a line for the `parser_args` parameter.

lib/krikri/parser.rb
```diff
@@ -22,8 +22,8 @@ def initialize(record)
     #
     # @param record the record to parse
     # @return [Krikri::Parser] a parsed record object
```
anarchivist: Need to add a comment line about the `args` parameter.

lib/krikri/mapping.rb
no-reply: Good call.

anarchivist:
Documentation nitpicking aside, I think this is ready to go.

no-reply:
I think that should do it.

PR #90, opened 2015-02-06
Mapper error logging & empty DSL URI

no-reply:
Adds error handling and logging for `Mapper` activities. These are handled similarly to `Harvester` error handling, this passes over failures on mapped records and logs the error message.

A separate commit addresses cases where empty value sets are passed to URI in the DSL.


anarchivist:
Looks good. Thanks! :white_check_mark:

no-reply:
:feelsgood:

PR #92, opened 2015-02-08
Streamline and improve Rake tasks

anarchivist:
* `rake ci` now invokes `engine_cart:clean` before running `engine_cart:generate`,
to allow for improved local testing by developers before pushing up their commits.
* Adds new `jetty:clean_marmotta_home` task, which deletes `MARMOTTA_HOME` if defined
to something other than the default. If `MARMOTTA_HOME` is set to the default, it will
automatically get removed by `jetty:clean`.
* Makes `jetty:clean_marmotta_home` a prerequsite for `jetty:clean`. This also removes a
second (accidental) invokation of `jetty:clean` from the previous tasks file.
* <strike>Adds a dependency on `rubyzip`, now required by Jettywrapper >= 2.0.1. Declaring
`rubyzip` as a development dependency was not adequate for some reason</strike> **Fixed by Jettywrapper >= 2.0.2**

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -25,7 +25,7 @@ require 'jettywrapper'
 import 'lib/tasks/jetty.rake'
 
 desc "Run all specs in spec directory (excluding plugin specs) in an engine_cart-generated app"
-task :ci => ['jetty:clean', 'engine_cart:generate'] do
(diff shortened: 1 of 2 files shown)


anarchivist:
This has the happy side effect of allowing development deployment using Jettywrapper on Windows machines because of projecthydra/jettywrapper#37 and projecthydra/jettywrapper#40.

PR #95, opened 2015-02-17
Add ability select multiple fields from ValueArray

no-reply:
Creates `ValueArray#fields` which accepts multiple fields/chains and gives their union.

Also adds:
  - `ValueArray#[]=` with type checking
  - `ValueArray#<<` with type checking
  - `ValueArray#flatten` to return a new `ValueArray`
  - `ValueArray#concat` to return a new `ValueArray`

### DIFF
--- a/lib/krikri/parser.rb
+++ b/lib/krikri/parser.rb
@@ -133,11 +133,36 @@ module Krikri
     class ValueArray
       include Enumerable
 
-      delegate :<<, :[], :[]=, :each, :empty?, :map, :to_a, :to_ary,
+      delegate :[], :each, :empty?, :map, :to_a, :to_ary,
       :to => :@array
 
       def initialize(array = [])
         @array = array
+      end
+
+      ##
+      # @see Array#[]=
+      # @raise [InvalidParserValueError] when the value is not a Parser::Value
+      def []=(index, value)
+        raise InvalidParserValueError unless value.is_a? Value
+        @array[index] = value
+        self
+      end
+
+      ##
+      # @see Array#<<
+      # @raise [InvalidParserValueError] when the value is not a Parser::Value
+      def <<(value)
+        raise InvalidParserValueError unless value.is_a? Value
+        @array << value
+        value
+      end
+
+      ##
+      # @see Array#concat
+      # @return [ValueArray]
+      def concat(*args, &block)
+        self.class.new(@array.concat(*args, &block))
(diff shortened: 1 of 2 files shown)


lib/krikri/parser.rb
```diff
@@ -216,6 +262,10 @@ def self.build(record)
       def get_field(name)
         self.class.new(flat_map { |val| val[name] })
       end
+
+      public
+
+      class InvalidParserValueError < StandardError; end
```
anarchivist: Should this subclass `TypeError`? That might be a little more appropriate, but I don't have a strong feeling either way.

lib/krikri/parser.rb
```diff
@@ -162,6 +187,18 @@ def field(*args)
       end
 
       ##
+      # Accesses the union of multiple specified fields.
+      #
+      # @return [ValueArray] an array containing the nodes available in the
+      # given fields.
+      def fields(*args)
```
anarchivist: If I'm reading this and the spec correctly, something like `records.fields(:foo, :bar)` (i.e, without explicit array typing) should return the union for the fields `:foo` and `:bar`. Correct?

anarchivist:
From Slack: 

```
< tom> i was also thinking about an interface like `record.field(:my, :field, :chain).and(:other, :field, :chain)`
< tom> that would take a slightly larger rewrite, but would allow things like
`record.field(:my, :field).select(&:meets_requirement?).and(:other, :field).reject(&:is_malformed?)`
```

I'd be :+1: on that, but I think that's OK as it's own story/PR. We can discuss tomorrow in standup.

anarchivist:
And, as it stands, I'm :+1: on this PR once we figure out how we want to deal with the two minor points I raised above.

lib/krikri/parser.rb
no-reply: Yes. 

And `records.fields([:foo, :bar], [:foo, :bar, :baz])` should return the union of the fields `:bar` in `:foo` and `:baz` in `:bar`.

anarchivist: :ok:

lib/krikri/parser.rb
no-reply: Agreed. 

Fixed and squashed. :construction:

anarchivist:
:cool: :sunglasses:

PR #96, opened 2015-02-18
Add `#header` node to OAI Parsers (and DSL method)

no-reply:
Adds a mixin module `OaiParserHeaders` for use with `XmlParser` classes.  This module adds a `#header` method which gives access to the `<oai:header>` tag and its members, even when that tag is out of scope of the root node.

Adds a specialized method to the mapping DSL via `ParserMethods` to access the header.  This method raises an error if no header exists.


anarchivist:
I'm :+1: :goat: :capricorn:  once we :squash: so we can :ship:.

no-reply:
:squash:'d

PR #97, opened 2015-02-19
Agent/Job refactors

no-reply:
Floating these minor changes as a simpler interface for `#enqueue`.

This makes it possible to do `MyHarvester(opts)` instead of `MyHarvester(HaverstJob, opts)`.  The initial parameter was previously passed to all harvesters.

Possible we could do better by building a way for `SoftwareAgent`s to specify their `Job` class?


anarchivist:
> Possible we could do better by building a way for `SoftwareAgent`s to specify their `Job` class?

Is this a now-refactor or a later-refactor?

no-reply:
> Is this a now-refactor or a later-refactor?

That's up for discussion.  If there's general agreement that this would be a good approach... now, I suppose.

There's maybe some discussion we could have about responsibilites, too. Currently:
  - `HarvestJob` has nice dependency injection: https://github.com/dpla/KriKri/blob/develop/lib/krikri/harvest_job.rb#L10
  - `MappingJob`, too: https://github.com/dpla/KriKri/blob/develop/lib/krikri/mapping_job.rb

It would be possible to collapse all of this into `Job` and do away with the whole issue, as long as we think the logic in `#run` can stay the same throughout. We would then do: `Resque.enqueue_to(self.queue_name, Krikri::Job, activity.id)`, here: https://github.com/dpla/KriKri/blob/develop/lib/krikri/software_agent.rb#L88

But maybe we don't want `SoftwareAgent` implementations to need to know their queue name?

no-reply:
This is no longer in the realm of pure refactor, since it changes the interface to `SoftwareAgent#enqueue`, but I think it tightens things up significantly.  Not only does it remove `#enqueue`'s responsibility to instantiate a specific `Job` subclass, it removes those subclasses altogether.

In the event that we need different `Job#run` logic for some jobs in the future, I think we could approach it with further decoupling.

anarchivist:
:+1:, whenever you're feeling happy with it. How much do you want to :jack_o_lantern:?

spec/lib/krikri/mapping_job_spec.rb
```diff
@@ -1,36 +0,0 @@
...
-    let(:mapping_activity) { create(:krikri_mapping_activity, opts: opts) }
-    let(:test_mapper) { instance_double(Krikri::Mapping) }
-
-    before do
-      record.save(activity.rdf_subject)
-      allow(Krikri::Mapper::Registry).to receive(:get).with(:test_map)
-        .and_return(test_mapper)
-    end
-
-    it 'calls the mapper with record' do
```
no-reply: This was a lot of setup to test something that is tested better elsewhere!  Glad to see this one go!

anarchivist: :clap: :recycle:

no-reply:
:squash:'d

spec/lib/krikri/harvester_spec.rb
```diff
@@ -17,6 +17,13 @@
     end
   end
 
+  describe '.enqueue' do
+    it 'defaults to HarvestJob' do
```
markbreedlove: Should this say "DummyHarvester"?

markbreedlove:
This is great to see that `Resque.enqueue_to` exists, because it means there's no need to have the derived classes of Job just to define the queue name.

> But maybe we don't want SoftwareAgent implementations to need to know their queue name?

My intuition says that's correct, and I think, was one of the reasons it wasn't declared there before.  I could see having different queues for different activities of the same SoftwareAgent.  A file-based harvester, for example, could want to be put in a different queue so that it runs on a particular server.  There could be other reasons that I can't think of now.

We can still do that without `HarvestJob`, `MappingJob`, etc., but we'll need to think of flexible queue names at some point, perhaps when we create provider ingestion profiles.

I'm assuming we're confident now that a `Job` would not have to do anything differently from agent to agent, such as pre- and post-tasks or logging; that this is better done inside of the agent when it's run.  It does seem cleaner that way, but I'm not sure that was totally clear months ago when the Job classes were originally written.

no-reply:
> It does seem cleaner that way, but I'm not sure that was totally clear months ago when the Job classes were originally written.

When the job classes were originally written, there were some minor differences in their `#perform` methods, which have since evaporated as the `SoftwareAgent` interface solidified.  In the event that interface differences start to re-emerge, I think we may be able to find a better pattern.

> I could see having different queues for different activities of the same SoftwareAgent. A file-based harvester, for example, could want to be put in a different queue so that it runs on a particular server.

Absolutely.  This is one reason why the option to pass a `queue_name` was added.  My sense is that the job subclasses had essentially become clunky aliases to queue names.  This frees us to create a specialized queue for an individual job, without pushing any code at all, which could come in handy.

PR #100, opened 2015-02-23
QA by provider

AudreyAltman:
This feature makes all existing QA tasks specific to a provider.  With this implementation, you cannot perform QA tasks on multiple provider data, nor can you perform QA tasks on the entirety of the saved/enhanced records.  You must choose one and only one provider.

The provider is stored as a session variable.  It is set when a user clicks on a provider link from the `admin#index` page.  The value for `session[:provider]` is currently the provider name, but could be changed to the provider id, or the institution id stored in the application's relational database.

The main methods for restricting QA tasks to one provider are in the module `Krikri::QaProviderFilter`.  This module can be mixed into controllers or libraries.

I also took the opportunity to refactor some of the business for querying the solr index into a new class, `Krikri::IndexQuerier`.  This makes things simpler and more consistent across `Krikri::ProviderList`, `Krikri::RandomRecordGenerator`, and `Krikri::ValidationReportList`.

There are two new controllers: `admin#index` will be a dashboard for administrator tasks.  `dashboards#index` will be a dashboard for hub tasks.


AudreyAltman:
I made a couple of changes to the views based on Gretchen's feedback.

config/routes.rb
```diff
@@ -9,5 +9,7 @@
   resources :institutions do
     resources :harvest_sources, shallow: true
   end
+  resources :admin, only: [:index]
```
anarchivist: I'm wondering if the root route that gets setup by the Krikri generator should point to `admin#index`. Thoughts?

lib/krikri/index_querier.rb
```diff
@@ -0,0 +1,17 @@
+module Krikri
+  # Queries the Search Index
+  class IndexQuerier
```
anarchivist: Can you talk a little more about how this refactor improves things? The name feels a little confusing. Is this a query builder? How does this all fit together?

lib/krikri/qa_provider_filter.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    ##
+    # Returns the :fq param that will filter a solr query by provider.
+    # @param String
+    # @return String
+    # @example:
+    #   query_params = { :fq => provider_fq("Boston Public Library") }
+    def provider_fq(provider)
+      "provider_name:\"#{provider}\""
```
anarchivist: Does `provider` need to be urlencoded, or does that happen automatically somewhere down the line?

lib/krikri/index_querier.rb
anarchivist: Does `Krikri::IndexQuerier` need tests?

lib/krikri/provider_list.rb
```diff
@@ -0,0 +1,19 @@
+module Krikri
+  # Contrtructs a list of providers
+  class ProviderList
+
+    def initialize
+      @index_querier = Krikri::IndexQuerier.new
```
no-reply: Maybe give the option to inject this?

```ruby
def initialize(index_querier = Krikri::IndexQuerier.new)
  @index_querier = index_querier
end
```

(I think I agree w/ @anarchivist that that class could use a better name).

lib/krikri/provider_list.rb
```diff
@@ -0,0 +1,19 @@
...
+      @default_params = { :rows => 10,
```
no-reply: Should this be a class-level constant, rather than an instance variable?

config/routes.rb
AudreyAltman: +1

lib/krikri/index_querier.rb
AudreyAltman: @anarchvist I created this class because I found that I kept repeating these three lines of code:

<pre>@repo = Blacklight::SolrRepository.new(Blacklight::Configuration.new
query_result = @repo.search(query_params)
Blacklight::SolrResponse.new(query_result, query_params)</pre>

We have three libraries so far that need to query the Solr index, and I believe there will be more.  Since we will eventually be moving to ElasticSearch, I am trying to keep the logic that deals with the SolrIndex as isolated as possible.  Maybe a class isn't best way to handle this.  I could just add this logic back into the individual libraries that use it, or I could make a module instead of a class.  I defer to your good judgement, and am happy to discuss further if it's not yet clear.

AudreyAltman: Re:testing, I am not sure what I could test that isn't just re-testing Blacklight functionality, so I would welcome suggestions.

lib/krikri/provider_list.rb
AudreyAltman: Sure - do you think I should also change `@default_params` to a class-level constant in `RandomRecordGenerator` and `ValidationReportList`?

lib/krikri/qa_provider_filter.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+  end
+end
```
AudreyAltman: @anarchivist and @no-reply I am not sure about the `QaProviderFilter` module.  It is included in both controllers and libraries, but two of its three methods, `records_by_provider` and `authenticate_session_provider` are specific to controllers.  Is it okay to have a single module that contains any logic that controllers or other classes might need to filter QA tasks by provider?  Let me know if this question makes sense.

lib/krikri/qa_provider_filter.rb
AudreyAltman: I don't see that it's urlencoded anywhere in the Blacklight code, so I'll add it here.

anarchivist: @AudreyAltman I'd check RSolr as well, because I get the sense that the encoding happens at that layer.

anarchivist: Just to close the loop - I think RSolr should handle this escaping automatically.

lib/krikri/qa_provider_filter.rb
anarchivist: Hmm. I don't have a strong feeling about this. I'm wondering if this should be placed in a concern module. @no-reply?

lib/krikri/index_querier.rb
anarchivist: Ah, I see what you're trying to do. At least for the first two, maybe we should address it this way: https://github.com/projectblacklight/blacklight/blob/master/lib/blacklight/search_helper.rb#L7-L16

anarchivist: I'm not sure if this addresses the need to return a new instance of `Blacklight::SolrResponse`, but it'll help with the first two calls probably?

markbreedlove:
Without reading anything beyond the main pull request comment up top, the first observation I have is that we're apparently using a session variable for the provider.  This may be fine, for now, but I'm usually wary of using session cookies because we can't really make a RESTful API for the QA app if we're relying on them.  (For a given resource, that returns something like a list of items for the provider, the provider ID is not conveyed in the URI of the request.)  This makes it more roundabout to write frontend Javascript to do things that are convenient for the users and reduce latency in updating parts of the u.i.

I don't necessarily know where I'm going with this, or how big an impact it has, and I'll have to read the code, but I just wanted to bring session variables up as something that can mean less-clean code if you want to do certain things with the u.i.

config/routes.rb
markbreedlove: +1

no-reply:
I think I'm with @markbreedlove that Provider ID would be better as a part of the request path or parameters.

lib/krikri/random_record_generator.rb
```diff
@@ -1,19 +1,36 @@
 module Krikri
   # Gets random records from the search index
   class RandomRecordGenerator
+    include Krikri::QaProviderFilter
 
     def initialize
-      @solr_repo = Blacklight::SolrRepository.new(Blacklight::Configuration.new)
+      @index_querier = Krikri::IndexQuerier.new
+      @default_query_params = { :id => '*:*',
```
markbreedlove: Does this make Solr search and return all of a provider's documents, and then randomly choose an offset from that list? If so, is there any way to point Solr directly at a random offset without retrieving a potentially large number of documents?

I think we do that with ElasticSearch somewhere, and I know it is done with CouchDB in the old system (not that CouchDB has anything to do with this, but it's a feature), and relational DBs.  I could also be reading this wrong.

anarchivist: The search matches all of a provider's documents, sorted randomly, and returns only one document.

markbreedlove: (I should have commented on the next line down.  It looks like it's pulling all documents, sorting by a random offset, and then returning the first row.  But I could be wrong, of course!)

markbreedlove:
I have just a few more observations at a high level.

The controllers named `AdminController` and `DashboardsController` break with the usual paradigm and don't represent the objects they seem to describe -- admins, dashboards.  There is a convention to have a `PagesController` for pages like the admin or dashboard page -- at least there was with CakePHP, so I assume it's the same with Rails.   This is useful for things that are landing pages, index pages, etc, that just serve as menus or as mostly static pages -- "mostly" because you can still have dynamic elements.  You can have the PagesController call various views depending on the path parameters it's given. The `AdminController` as it exists is basically a list of providers, so it should probably be `ProvidersController`.

Also, with the MVC paradigm:  there are various classes now in the `lib` directory that are doing what models do, e.g. returning lists of things, given some parameters.  Do you think validation reports could be represented with a model?  When `ValidationReportList` makes a query on the search index to get a list of report links, it's doing a database query the same as an ActiveModel model for a relational database, except the database happens to be Solr under the hood.

I'm sorry if some of this is out of context, because I don't recall commenting on the original pull request that introduced the validation reports.

Also, I know there's a need to get this merged soon, so I don't see anything that should necessarily hold that up, but I'm just pointing out things that we may want to revisit.

AudreyAltman:
Re: using a session variable for providers.  My rationale for doing this was mainly so that an admin could access parts of the website as if she were logged in as a provider.  This includes not only accessing that provider's dashboard and completing solr queries by that provider, but also things like seeing the provider's name in the navigation bar.  To me, it fits the paradigm of a session state.  Also, since any non-admin logged in user is going to be associated with a provider institution, the identity of the provider is going to be somehow saved in a session variable anyway, no?  I am not opposed to changing the code so that the provider ID is in the URI rather than a session variable, since I understand that it's more RESTful that way, but I thought it worth a quick conversation

no-reply:
This makes sense, @AudreyAltman.  I think maybe there's a middle ground.  The guideline I would try to follow is that things like the navigation bar and permissions depend on session state, while access to "resources" (web pages, reports, provider-specific searches) is URI driven.

Does that make sense?

markbreedlove:
The main complaint versus session cookies, or any cookies, from the REST point of view is that they prevent the transactions from being stateless.  Cookies have been described as breaking the client's back button.

I know some partners might only look at one provider page, but others, like administrators, might want to jump around between different provider pages.  Also, sharing information about a given resource between frontend and backend code is more cumbersome if the frontend code can't analyze the URI to get everything it needs.

Ideally we would not be drawing the whole HTML page every time something on the page needs to be updated; though I know it's easy to do that with any MVC framework like Rails.  We should have endpoints that enable us to redraw small sections of the page dynamically with frontend code. (Later, not right away because I acknowledge things need to just get done :-) )

I think that, ideally, there would be one authentication endpoint that could accept a session cookie, but that returns an access token, which then gets put into the URL for every request. (That may be doable with what's there now with Devise.)   Anything like the provider could go into the URL.  Then when you go through your browser history tomorrow and try to go back to provider page x, it either just works or, at worst, you get redirected to a login page, which is then capable of redirecting you back to the page you wanted by virtue of the fact that the application state is conveyed entirely in the URL.  If a crucial session cookie were expired, that had the provider (or a browser cookie, where that state were stored in the client) it would be less smooth -- someone would have to presumably log in and then navigate around, and the bookmark or history item would not have been useful.

lib/krikri/index_querier.rb
AudreyAltman: After a conversation with Tom, I am refactoring this class and renaming it `SolrResponseBuilder`.

no-reply:
Closing this, since #134 is now. :rocket:

PR #101, opened 2015-02-24
Set fragment URI for first sourceResource on ore:Aggregations

anarchivist:
cc @no-reply - perhaps you can give this a sanity check, or help out on this branch?


no-reply:
I think this should do it for you.

PR #103, opened 2015-02-25
Use solr settings in `IndexService`

no-reply:
This makes it possible to configure the `IndexService` to use a solr server other than `localhost`.

Testing this is hard because `IndexService` is effectively a singleton... thinking about a refactor.

### DIFF
--- a/lib/krikri/engine.rb
+++ b/lib/krikri/engine.rb
@@ -15,6 +15,12 @@ module Krikri
   class Engine < ::Rails::Engine
     isolate_namespace Krikri
 
+    def configure_blacklight!
+      krikri_solr = Krikri::Settings.solr
+      Blacklight.solr_config = Blacklight.solr_config.merge(krikri_solr) unless
+        krikri_solr.nil?
+    end
+
     # Autoload various classes and modules in lib
     config.autoload_paths += Dir["#{config.root}/lib/**/"]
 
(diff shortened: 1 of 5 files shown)


no-reply:
Updated to use instances of the class.  This will break current users of the `IndexService` (krikri tasks).  These still need to be updated before this is ready to merge.

no-reply:
We need a strategy for testing our rake tasks.  In the meanwhile, I would encourage reviewers to run them manually before :+1:'ing this.

no-reply:
Need to share configuration with blacklight.  Proposal is to monkeypatch: https://github.com/projectblacklight/blacklight/blob/master/lib/blacklight.rb#L78

no-reply:
I think the approach taken is much better than the monkeypatch idea (thanks @anarchivist) and this is ready for review.

markbreedlove:
For the record, I've done a test to sanity-check this.  I modified `config/settings/test.yml` with the following configuration, which causes the the spec test `spec/lib/krikri/index_service_spec.rb:69` to fail because it has a bogus Solr URL:

```
solr:
    url: 'http://nogood/'
```

If I take that out, the `test` configuration in the `solr.yml` file takes effect, and the test passes.
This just verifies that this `merge()` works as planned to let the setting be overridden:  https://github.com/dpla/KriKri/commit/3e87e0109f5ce9fd97f7a3cf3db461884ebaa4a4#diff-b31c2deb469ce65b5f99d9b33041a7c9R70

no-reply:
@markbreedlove sounds like your test verifies: https://github.com/dpla/KriKri/pull/103/files#diff-97e16f3fe7583a30f6ac1a80a9c8bae6R16 ?

are we good to :squash: ?

markbreedlove:
Yes, looks good :+1:

anarchivist:
:ok_hand: on my end too.

PR #104, opened 2015-02-26
Rework `#header` parser method; add integration test

no-reply:
Okay... integration testing... blah.

### DIFF
--- a/lib/krikri/mapping_dsl/parser_methods.rb
+++ b/lib/krikri/mapping_dsl/parser_methods.rb
@@ -48,11 +48,10 @@ module Krikri::MappingDSL
     #   OriginalRecord associated with the parsed record passed as its argument.
     # @todo consider a more generalized approach
     def header
-      lambda do |parsed|
-        raise "#{parsed} does not have a `header`" unless
-          parsed.respond_to? :header
-        parsed.header
-      end
+      RecordProxy.new(
+        [{ name: :header }],
+        Krikri::Parser::ValueArray,
+        true)
     end
 
     ##
@@ -65,7 +64,7 @@ module Krikri::MappingDSL
     #
     # @see Krikri::Parser::ValueArray
     class RecordProxy
-      attr_reader :value_class, :call_chain
+      attr_reader :value_class, :call_chain, :non_root
 
       ##
       # Create a new RecordProxy object.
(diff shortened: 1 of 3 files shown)


lib/krikri/mapping_dsl/parser_methods.rb
```diff
@@ -78,9 +77,10 @@ class RecordProxy
       #   Krikri::Parser::ValueArray
       #
       # @return [RecordProxy]
-      def initialize(call_chain = [], klass = Krikri::Parser::ValueArray)
+      def initialize(call_chain = [], klass = Krikri::Parser::ValueArray, non_root = false)
```
anarchivist: as a new parameter, `non_root` should be documented

anarchivist:
I'm testing this against ESDN, and I'm still struggling with it...

The following:

```ruby 
    collection :class => DPLA::MAP::Collection, :each => header.field('xmlns:set_spec'), :as => :coll do
      title coll
    end
```

fails with:

```
undefined method `header' for #<Krikri::Parser::ValueArray:0x007fa3bdff33a0>
/Users/mark/Documents/Code/dpla/krikri/lib/krikri/mapping_dsl/parser_methods.rb:101:in `block in call'
/Users/mark/Documents/Code/dpla/krikri/lib/krikri/mapping_dsl/parser_methods.rb:100:in `each'
/Users/mark/Documents/Code/dpla/krikri/lib/krikri/mapping_dsl/parser_methods.rb:100:in `call'
/Users/mark/Documents/Code/dpla/krikri/lib/krikri/mapping_dsl/property_declaration.rb:32:in `block in to_proc'
```

---

The following:

```ruby 
    collection :class => DPLA::MAP::Collection, :each => record.header.field('xmlns:set_spec'), :as => :coll do
      title coll
    end
```

fails with:

```
undefined method `header' for #<Krikri::MappingDSL::ParserMethods::RecordProxy:0x007f9d307c1800>
/Users/mark/Documents/Code/dpla/krikri/lib/krikri/mapping_dsl/parser_methods.rb:122:in `method_missing'
/Users/mark/Documents/Code/dpla/heidrun/vendor/mappings/heidrun-mappings/heidrun/esdn_mods.rb:28:in `block (2 levels) in <top (required)>'
/Users/mark/Documents/Code/dpla/krikri/lib/krikri/mapping_dsl/child_declaration.rb:33:in `instance_eval'
/Users/mark/Documents/Code/dpla/krikri/lib/krikri/mapping_dsl/child_declaration.rb:33:in `block in to_proc'
```

no-reply:
Was missing the new parameter in `#dup` which was causing a problem calling within child declarations.

This should now be ready for review.

anarchivist:
:squash: :ok_hand:

PR #105, opened 2015-02-27
Forceably update SourceResource instances on Aggregations

anarchivist:
This extends upon dpla/krikri#101, which did not fully fix the issue. It turns out
there is an underlying issue with ActiveTriples (see ActiveTriples/ActiveTriples#107).
This rebuilds the SourceResource and appropriately sets the subject to the required
fragment URI.

Do not merge until the tests are done.

### DIFF
--- a/lib/krikri/engine.rb
+++ b/lib/krikri/engine.rb
@@ -82,8 +82,25 @@ module Krikri
 
         def mint_id!(seed = nil)
           set_subject!(mint_id_fragment(seed))
-          sourceResource.first.set_subject!(rdf_subject / '#sourceResource') \
-            unless sourceResource.empty?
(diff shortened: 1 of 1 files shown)


anarchivist:
I think this is ready for review now; let me know if you think this needs better testing.

no-reply:
:squash: I'm satisfied with this for now.  I'm working on it as a part of the `ActiveTriples` persistence strategy work.  Lots of progress on that in the last few days, and I'm thinking this can hold us over until `ActiveTriples` is a bit closer to 1.x.

https://github.com/ActiveTriples/ActiveTriples/compare/mixin-resource...feature/persistence-strategies

anarchivist:
Squashed; :merge: :question:

PR #106, opened 2015-02-27
Add set access, multiple harvests and blacklisting to OAIHarvester

no-reply:
- Adds `OAIHarvester#sets` to list sets.
 - Adds ability to specify multiple sets to harvest by passing a list to the `:set` option on `#records` and `#record_ids`.
 - Adds ability to blacklist sets in `#records` and `#record_ids` requests.

### DIFF
--- a/lib/krikri/harvesters/oai_harvester.rb
+++ b/lib/krikri/harvesters/oai_harvester.rb
@@ -60,13 +60,34 @@ module Krikri::Harvesters
       end
     end
 
-    # TODO: normalize records; there will be differences in XML
-    # for different requests
+    ##
+    # Gets a single record with the given identifier from the OAI endpoint
+    #
+    # @param identifier [#to_s] the identifier of the record to get
+    # @param opts [Hash] options to pass to the OAI client
     def get_record(identifier, opts = {})
       opts[:identifier] = identifier
       opts = opts.merge(@opts)
       @record_class.build(mint_id(identifier),
                           record_xml(client.get_record(opts).record))
+    end
+
+    ##
+    # Lists the sets available from the OAI endpoint. Accepts a block to
+    # pass to `#map` on the resulting array.
+    #
+    # @example:
+    #
+    #   sets(&:spec)
+    #
+    # @param opts [Hash] options to pass to the OAI client
+    # @return [Array<OAI::Set>] an array of sets.
+    #
+    # @see OAI::Set
+    def sets(opts = {}, &block)
+      arry = client.list_sets.full.to_a
+      return arry unless block_given?
+      arry.map(&block)
     end
 
     ##
@@ -76,7 +97,9 @@ module Krikri::Harvesters
         key: :oai,
         opts: {
           set: {type: :string, required: false, multiple_ok: true},
-          metadata_prefix: {type: :string, required: true}
+          metadata_prefix: {type: :string, required: false},
+          from: {type: :string, required: false},
+          until: {type: :string, required: false}
         }
       }
     end
--- a/lib/krikri/harvesters/oai_harvester.rb
+++ b/lib/krikri/harvesters/oai_harvester.rb
@@ -1,6 +1,29 @@
 module Krikri::Harvesters
   ##
   # A harvester implementation for OAI-PMH
+  #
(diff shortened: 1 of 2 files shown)


lib/krikri/harvesters/oai_harvester.rb
```diff
@@ -1,6 +1,27 @@
...
+  #          If multiple sets are given, they will be lazily requested from
+  #          `OAI::Client#list_records` in turn and combined into a single
+  #          enumerator.
+  #   - skip_set: A string or array of strings specifying the sets to skip.
+  #               If both `set` and `skip_set` are given, sets specified as
+  #               skip are excluded from the harvest. Otherwise, all sets
+  #               returned by `#set` except those skipped will be harvested.
+  #   - metadata_prefix: A string specifying the metadata prefix. e.g. 'oai_dc'.
+  #   - from: The begin date for the harvest.
+  #   - until: The end date for the harvest.
```
anarchivist: Should there be a `@see` reference to the OAI gem docs?

lib/krikri/harvesters/oai_harvester.rb
```diff
@@ -51,38 +76,107 @@ def count
...
-                            record_xml(rec))
-
+      opts = @opts.merge(opts)
+      request_with_sets(opts) do |set_opts|
+        client.list_records(set_opts).full.lazy.flat_map do |rec|
+          @record_class.build(mint_id(rec.header.identifier),
+                              record_xml(rec))
+        end
       end
+      # sets = Array(opts.delete(:set) { [nil] })
```
anarchivist: Do you still want to retain this commented-out code for some reason?

lib/krikri/harvesters/oai_harvester.rb
```diff
@@ -51,38 +76,107 @@ def count
...
+          until: {type: :string, required: false}
         }
       }
     end
 
+    ##
+    # Concatinates two enumerators
+    # TODO: find a better home for this. Reopen Enumerable?
+    def concat_enum(*enums)
+      Enumerator.new do |yielder|
```
anarchivist: Would it be possible to do something like `enums.flat_map { |e| e }`, or am I missing something about the seeming simplicity of that?

spec/lib/krikri/harvesters/oai_harvester_spec.rb
```diff
@@ -243,6 +378,20 @@
       end
     end
 
+    describe 'concat_enum' do
+
+      it 'concatenaces enums' do
```
anarchivist: typo

anarchivist:
I think this looks good. My most substantive feedback is about `#concat_enums`...

lib/krikri/harvesters/oai_harvester.rb
no-reply: I think `enums.flat_map { |e| e }` is a no-op.  I thought `enums.flat_map { |e| e.each { |i| i } }` might work, but that seems also to fail nastily.

spec/lib/krikri/harvesters/oai_harvester_spec.rb
no-reply: what excellent spelling you have, @no_reply

no-reply:
I believe this is ready. I'm still not singing the praises of `#concat_enums`... or it living in `OAIHarvester` (or really, our code at all... why isn't this in Ruby?) but it works, and I'm inclined to run with it.

I can move it to its own `ConcatenatedEnumerator` class, if that's desired.

lib/krikri/harvesters/oai_harvester.rb
no-reply: Done.

anarchivist:
I don't know if we want to rely on an external gem to do this, but there's also `Enumerating`: https://github.com/mdub/enumerating#lazy-combination-of-enumerables

anarchivist:
But otherwise, I'm good on :squash:ing.

no-reply:
Yeah, I was reluctant to introduce a (unmaintained?) dependency for something so small.  I'm going to add a comment with a reference to that, so if we end up doing more of this concatenation, we can reevaluate.

This is squashed to the level I think is appropriate.

PR #107, opened 2015-03-03
Feature/skip deleted

no-reply:
- Add harvest behavior objects to harvester. This makes it possible to introduce new record processing behavior by injecting them into the harvesters.
- Create skip deleted behavior for OAI, make it the default behavior for OAI harvesters.

### DIFF
--- a/lib/krikri/harvester.rb
+++ b/lib/krikri/harvester.rb
@@ -22,14 +22,14 @@ module Krikri
   #      recommended to override this method in your subclass with
   #      an efficient implementation.
   #    - #run. Wraps persistence of each record returned by #records.
-  #      Runs a full harvest, inserting Original Records into the database,
-  #      given the options passed to #initialize.
+  #      Runs a full harvest, processing the created `record_class`
+  #      instances through `harvest_behavior`, given the options passed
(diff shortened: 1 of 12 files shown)


markbreedlove:
@no-reply and I have discussed this, and I think it looks good to merge.

I am about to make some compatible changes related to indexing and mapping activities, and will have a PR ready soon.

We might change where the behavior classes are located, but we're going to defer doing so until both of our branches are merged.

lib/krikri/harvester.rb
```diff
@@ -165,6 +177,18 @@ def self.expected_opts
       raise NotImplementedError
     end
 
+    protected
```
anarchivist: Do we still need to retain this commented-out protected method?

anarchivist:
I'm :+1: on this; just curious if we're planning to retain the commented method. Fix up and :squash: as appropriate.

no-reply:
I can fix up and squash once I'm in the air.

anarchivist:
:airplane: :jack_o_lantern:

no-reply:
I think this is ready. :rocket:

anarchivist:
:100:

PR #113, opened 2015-03-09
Add migration to change Krikri::Activity.opts from :string to :text

anarchivist:
Embiggens the `:opts` JSON blob used to pass options, to ensure cromulent option
storage and behavior for `Krikri::Activity` objects.

### DIFF
--- a/db/migrate/20150309160112_change_opts_in_krikri_activities.rb
+++ b/db/migrate/20150309160112_change_opts_in_krikri_activities.rb
@@ -1,0 +1,5 @@
+class ChangeOptsInKrikriActivities < ActiveRecord::Migration
+  def change
+    change_column :krikri_activities, :opts, :text
+  end
+end
--- a/spec/factories/krikri_activities.rb
+++ b/spec/factories/krikri_activities.rb
@@ -4,6 +4,14 @@ FactoryGirl.define do
     opts '{"uri": "http://example.org/endpoint"}'
   end
 
+  factory :krikri_activity_with_long_opts, parent: :krikri_activity do
+    opts '{"uri": "http://example.org/endpoint",' \
+      '"oai": {"metadata_prefix": "mods", "set": ["SSDPLABrynMawr",' \
+      '"SSDPLACornell","SSDPLAUCSD","SSDPLAWashington","SSDelwareAtlas",' \
+      '"SSDelwareGeorge","SSDelwareHistoric","SSDelwareIncorporated1968",' \
+      '"SSDelwareIncorporated1959"]}}'
+  end
+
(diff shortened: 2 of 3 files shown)


no-reply:
Does `krikri_harvest_activity` actually get saved at some point in the test suite?  (Does the factory linter take care of this?)

anarchivist:
I'm not sure, but my understanding is that `FactoryGirl.create` instantiates a "saved" instance. Whether that instance is actually persisted, I'm not sure.

It's also worth noting that from the standpoint of testing, ActiveRecord's implementation of column types on SQLite 3 means that `:string`: and `:text` get treated basically the same. `:string` is mapped to `VARCHAR` in SQLite, which has no explicit max length unique to it's type, while in ActiveRecord 4.1.9, in PostgreSQL it's mapped to `character varying(255)`.

anarchivist:
To clarify, while the SQLite 3 adapter in ActiveRecord 4.1.9 declares the [`:string` type to be `VARCHAR(255)`](https://github.com/rails/rails/blob/v4.1.9/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb), the SQLite documentation states that any field length given for a [`VARCHAR` is silently ignored](https://www.sqlite.org/datatype3.html) (see "2.2 Affinity Name Examples").

no-reply:
> I'm not sure, but my understanding is that FactoryGirl.create instantiates a "saved" instance. Whether that instance is actually persisted, I'm not sure.

It does and is, but I'm not seeing it called anywhere. Sorry if I'm missing it.

Given your other comments, maybe we should consider PostgreSQL in CI?

anarchivist:
>> @anarchivist: I'm not sure, but my understanding is that FactoryGirl.create instantiates a "saved" instance. Whether that instance is actually persisted, I'm not sure.
>
> @no-reply: It does and is, but I'm not seeing it called anywhere. Sorry if I'm missing it.

[Line 15 of `spec/models/activity_spec.rb`](https://github.com/dpla/KriKri/blob/develop/spec/models/activity_spec.rb#L15).

anarchivist:
Also, re PostgreSQL in CI: yes, we should probably run it against both SQLite and PG, and maybe MySQL too. I don't know if that's a priority for right now.

no-reply:
> Line 15 of spec/models/activity_spec.rb.

That's a call to the parent, not to `krikri_harvest_activity`, yes?

anarchivist:
OK, this is now explicitly running a test on `#save` on large `:opts` values. Let me know when you'd like me to :squash:.

no-reply:
This looks ready to me. :squash:

anarchivist:
:mans_shoe:
:bug:

PR #114, opened 2015-03-09
Make OAI multiset harvest truly lazy.

no-reply:
OAI harvests were sending their initial request for each set of options when the `#records` or `#record_ids` method was called. This meant that multiset harvests took a long time to start and were at risk of having their resumption tokens timed out. This fixes the calls to be more lazy,
so the initial OAI request for each call is delayed until those records are actually needed.


anarchivist:
:+1: :zzz:

PR #115, opened 2015-03-09
Add `#map` to DSL Parser::ValueArray methods

no-reply:
This allows you to call `#map` on DSL values to implement transforms.


no-reply:
Hold off on this.  I'm updating it.

PR #116, opened 2015-03-09
Feature/index an activity

markbreedlove:
** DO NOT MERGE **

This is just for comment.  I am not done with this branch yet.

I've added "behavior" classes for activities, in preparation for the addition of indexing activities, and having activities report on the entities they generated.

There are some FIXME and TODO items that need attention:

1. My [factory method for representing an enrichment](https://github.com/dpla/KriKri/blob/4f3012063bc0b17626f1f90b3ad8331c57a17bca/spec/factories/krikri_activities.rb#L25-L62) is not working
2. There's an [issue with representing an options hash](https://github.com/dpla/KriKri/blob/4f3012063bc0b17626f1f90b3ad8331c57a17bca/spec/lib/krikri/mapper_agent_spec.rb#L9-L33) in the Mapper::Agent tests.
3. I want to remove [`Mapper::Agent#target_records`](https://github.com/dpla/KriKri/blob/feature/index-an-activity/lib/krikri/mapper.rb#L135-L164) (formerly named `records`) but I'm having trouble refactoring the spec tests to get that to happen.

Next up, also for comment:
- I think the behavior classes should be moved into a subdirectory
- I need to add a test for SoftwareAgent#set_generator_activity
- I need to actually add Elasticsearch indexing!


lib/krikri/software_agent.rb
```diff
@@ -30,6 +30,38 @@ def log(priority, msg)
...
+    # return the class of the appropriate behavior.
+    #
+    # We look for either `generator_activity' or `generator_uri' in the `opts'
+    # keys.  `generator_activity' must be a Krikri::Activity object, and
+    # `generator_uri' can be a string or RDF::URI
+    #
+    # @see Krikri::Mapper::Agent
+    # @see Krikri::Harvester
+    #
+    def set_generator_activity!(opts)
```
no-reply: Can we just clean this up, choose an option name/format and move forward.  This is a lot of code/validation for something that seems like ti could be really simple.

no-reply: I'm also wondering whether this should live in `SoftwareAgent` at all.  For many agents (harvesters, for example) , this won't come into play.

lib/krikri/harvester.rb
```diff
@@ -43,6 +43,12 @@ class << self
         def queue_name
           :harvest
         end
+        def entity_behavior
+          Krikri::OriginalRecordEntityBehavior
+        end
+        def generator_entity_behavior
+          raise 'A Harvester does not have a generator Activity.'
```
no-reply: This is a major code smell, to me.

See: https://github.com/dpla/KriKri/pull/116/files#r26081301

lib/krikri/mapper.rb
```diff
@@ -90,23 +90,55 @@ def mapping_exception_desc(rec)
...
 
       def self.queue_name
         :mapping
       end
 
+      def self.entity_behavior
+        Krikri::AggregationEntityBehavior
+      end
+
+      def self.generator_entity_behavior
```
no-reply: Make these instance methods. Otherwise you need to subclass the `Agent` to get a different behavior, which seems to defeat the purpose of injecting them.

spec/support/shared_contexts/generated_entities_query.rb
```diff
@@ -0,0 +1,18 @@
...
+  # aggregation comes from lib/dpla/map/factories.rb in DPLA::MAP
+  let(:aggregation) { build(:aggregation) }
+
+  before do
+    allow(DPLA::MAP::Aggregation).to receive(:new)
+      .with(solution.record.to_s)
+      .and_return(aggregation)
+    allow(aggregation).to receive(:get).and_return(true)
+  end
+end
```
no-reply: newline?

spec/factories/krikri_activities.rb
```diff
@@ -1,18 +1,65 @@
...
+  # [9] pry(main)> Krikri::Mapper::Agent.include?(Krikri::SoftwareAgent)
+  # => true
+  # [10] pry(main)> Krikri::Enrichments::StripHtml.include?(Krikri::SoftwareAgent)
+  # => false
+  # [11] pry(main)> e = Krikri::Enrichments::StripHtml.new
+  # => #<Krikri::Enrichments::StripHtml:0x007fa08ef3c100>
+  # [12] pry(main)> e.respond_to?(:agent_name)
+  # => false
+  #
+  # It's not clear from the inline enrichment comments how to enqueue
```
no-reply: This should be considered well out of scope for this PR (there's a story for it here: https://www.pivotaltracker.com/story/show/82567804).

In short, Enrichments were originally envisioned as implementing the `SoftwareAgent` interface, but that was before the Agent encapsulated queuing logic.  At this point, that relationship should probably be removed.

Some other Agent will be introduced (`EnrichmentProfile`?) when we move past transaction script style enrichments.

no-reply:
I've offered the feedback I think I can, at this point.  The bulk of this is a big :+1:.

There are a few places where I think there's some more ironing out to do, and I've called those out in the inline comments.

Other than that, I'd add that it would be great to see more mature test coverage of this, at this stage of development.

lib/krikri/software_agent.rb
markbreedlove: This is here because we used to only be passed a generator URI when instantiating an agent, but now the agent needs a property representing an activity; because that has a behavior that we need, which tells us the entities that were generated.

`initialize` was always passed `generator_uri`, for mappers, at least, but we need an activity object.

If an agent like a harvester doesn't have a generator URI, this does nothing.

Sure, it could be trimmed down by removing the handling of `generator_activity`, but I probably put that there because I could imagine a good use for it.

no-reply: Yeah, I think I wasn't very clear.  I'm :+1: on all the generator handling, I'm just advocating for making it a separate mixin or set of composable functionality rather than sitting in the base `SoftwareAgent`, since `SoftwareAgent`s don't necessarily work with records according to `Activity`.

spec/factories/krikri_activities.rb
markbreedlove: I could have made this more clear, but I am not suggesting changing this in this branch.

I am suggesting that it gets addressed somewhere else so that I can rebase upon it.  The fact is we have enrichment classes that appear to be software agents, that can not be represented in this factory, and it just makes sense to have them accounted for here if we're already doing harvests and mappings.

Following the example of the existing `mapper_agent_spec.rb`, which was written to verify that a mapper agent can get the records generated by a harvest activity, it only made sense to eventually write an enrichment agent that gets the records generated by a mapping activity, and an indexing agent that gets the records generated by an enrichment activity.

lib/krikri/mapper.rb
markbreedlove: I was simply taking a cue from `Mapper::Agent.queue_name`.  Why would that be a class method, but these not be?  Why would we ever want to say that a mapper does not generate aggregations?

no-reply: This is a good point. The main difference seems to be that `queue_name` is used from the class method `.enqueue`. It can eventually be overridden at the method call level (https://github.com/dpla/KriKri/blob/develop/lib/krikri/software_agent.rb#L95).

> Why would we ever want to say that a mapper does not generate aggregations?

Mappers accept a model class on definition, with the idea of supporting conversion to non-DPLA MAP models: https://github.com/dpla/KriKri/blob/feature/index-an-activity/lib/krikri/mapper.rb#L41 

It might be neat to have the Agent infer its behavior (or maybe just the class to build) from the mapper itself.

spec/factories/krikri_activities.rb
no-reply: > The fact is we have enrichment classes that appear to be software agents, that can not be represented in this factory...

This seems like the problem to me.  The Enrichments mixin `SoftwareAgent`, but don't  (and shouldn't, I think) implement any of its interface. 

A better way to put my original comment might have been:

> Good catch. Enrichments are hopelessly confused with respect to `SoftwareAgent` implementation (mea culpa).  I think the best approach might be to treat them as non Agents for the purpose of this PR, and push to sort this out as a part of https://www.pivotaltracker.com/story/show/82567804

app/models/krikri/activity.rb
```diff
@@ -83,5 +83,23 @@ def rdf_subject
       RDF::URI(Krikri::Settings['marmotta']['ldp']) /
         Krikri::Settings['prov']['activity'] / id.to_s
     end
+
+    ##
+    # Return an Enumerator of URI strings of entities (e.g. aggregations or
+    # original records) that pertain to this activity
+    #
+    # @return [Enumerator] URI strings
```
anarchivist: Minor: This should state it returns an `Enumerator::Lazy`.

anarchivist: Err, never mind. Was looking at the wrong line.

lib/krikri/aggregation_entity_behavior.rb
```diff
@@ -0,0 +1,14 @@
+
+# TODO:  move to a subdirectory?
```
anarchivist: Maybe `lib/krikri/entity_behaviors`, as we've done with something like parsers? (see [here](https://github.com/dpla/KriKri/blob/develop/lib/krikri/parsers/json_parser.rb))

lib/krikri/aggregation_entity_behavior.rb
```diff
@@ -0,0 +1,14 @@
...
+
+module Krikri
+  class AggregationEntityBehavior < EntityBehavior
```
anarchivist: Should the parent class be namespaced `Krikri::EntityBehavior`?

lib/krikri/mapper.rb
anarchivist: > It might be neat to have the Agent infer its behavior (or maybe just the class to build) from the mapper itself.

:+1:

lib/krikri/aggregation_entity_behavior.rb
markbreedlove: I was taking a cue from some of the behaviors in the `harvesters` directory, where there aren't namespaces, like `OAISkipDeletedBehavior < BasicSaveBehavior` but I am happy to add the namespace.

anarchivist: I'm not sure, but the name spacing may be necessary if we arrange these subclasses into their own subdirectory, as I suggest.

lib/krikri/mapper.rb
markbreedlove: I'm realizing now that this is the reason I had it be a class method:
https://github.com/dpla/KriKri/blob/feature/index-an-activity/app/models/krikri/activity.rb#L101

`agent` is a string field of `Activity`, from the database record that's saved when the job is run.  I need to be able to determine the behavior associated with that agent's class.  I'll ponder alternative ways of doing it, but it's not obvious to me yet how to do it without it being a class method.

no-reply: Would `Activity#agent_instance` work?

lib/krikri/aggregation_entity_behavior.rb
markbreedlove: Yes, that seems best.  I was thinking about a general `behaviors` directory, but that would be out of scope here, and would not make all of the other classes easier to find.

markbreedlove:
Thanks for your help on this, @anarchivist and @no-reply.  I'm closing this and will issue another PR later.

PR #117, opened 2015-03-09
Add `#map` to DSL Parser::ValueArray methods 

no-reply:
This allows you to call `#map` on DSL values to implement transforms.


anarchivist:
:ok_hand: Ready to :squash:?

no-reply:
:squash:'ed

anarchivist:
:cake:

PR #119, opened 2015-03-10
use local names for item ids in routes

AudreyAltman:
This fixes a bug in which items with dots in their ids create invalid routes.  Instead of using the full id uri, routes instead use local names.

### DIFF
--- a/app/controllers/krikri/records_controller.rb
+++ b/app/controllers/krikri/records_controller.rb
@@ -80,5 +80,19 @@ module Krikri
 
       config.solr_document_model = Krikri::SearchIndexDocument
     end
+
+    ##
+    # Construct a valid item URI from a local name, and use it to fetch a single
+    # document from the search index.
+    # Override method in Blacklight::SolrHelper.
+    # TODO: This method is depreciated in Blacklight v5.10.
+    # TODO: Write appropriate test for this functionality after it is updated
+    # with Blacklight v5.10.
+    # @param String id is a local name.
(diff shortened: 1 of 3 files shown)


app/controllers/krikri/records_controller.rb
```diff
@@ -80,5 +80,17 @@ class RecordsController < CatalogController
...
       config.solr_document_model = Krikri::SearchIndexDocument
     end
+
+    ##
+    # Construct a valid item URI from a local name, and use it to fetch a single
+    # document from the search index.
+    # Override method in Blacklight::SolrHelper.
+    # This method is depreciated in Blacklight v5.10.
+    # @param String id is a local name.
+    def get_solr_response_for_doc_id(id=nil, extra_controller_params={})
```
no-reply: Do we want a test for this?

no-reply:
I'm :+1: on this once Travis finally builds it, and with the caveat of discussion about the need for a unit test on `get_solr_response_for_doc_id`.

app/controllers/krikri/records_controller.rb
AudreyAltman: As per slack conversation, I added `#TODO` comment in the controller indicating that an appropriate test should be written once we upgrade to Blacklight v5.10+.

no-reply: Awesome. I'll build this locally and try to merge in time to sneak it into the next Heidrun release.

anarchivist:
:+1:

PR #120, opened 2015-03-11
Add test coverage with simplecov

markbreedlove:
Here's a quick little addition for test coverage.

I added this quickly and it seems to work well enough; gives html reports in `./coverage`

It shouldn't run in ci tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ marmotta
 
 ./spec/internal
 dump.rdb
+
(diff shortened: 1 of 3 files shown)


anarchivist:
Great, thanks @markbreedlove! Just FYI, CodeClimate also runs code coverage reports for us automatically (in CI): https://codeclimate.com/github/dpla/KriKri

PR #121, opened 2015-03-11
Add controller/view for Krikri::Activity; add PrettifierHelperBehavior

anarchivist:
This was a self-serving thing I built out to make it easier for me to find and inspect instances of `Krikri::Activity` as I wrote the SPARQL queries for QA.

This also moves two methods from `Krikri::SearchResultsHelperBehavior` into a new
module, `Krikri::PrettifierHelperBehavior`.

### DIFF
--- a/lib/krikri/prettifier_helper_behavior.rb
+++ b/lib/krikri/prettifier_helper_behavior.rb
@@ -1,0 +1,19 @@
+module Krikri
+  module PrettifierHelperBehavior
+    def prettify_json_string(string)
+      begin
+        return JSON.pretty_generate(JSON.parse(string))
+      rescue JSON::ParserError
+        return string
+      end
+    end
+
+    def prettify_xml_string(string)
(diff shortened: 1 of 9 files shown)


lib/krikri/prettifier_helper_behavior.rb
```diff
@@ -0,0 +1,19 @@
+module Krikri
+  module PrettifierHelperBehavior
```
no-reply: :+1:

no-reply:
I'm :+1: on this. Anyone else want to give this a quick review (ping: @AudreyAltman)?  If not, I'll merge it so we can get it into the next release.

lib/krikri/prettifier_helper_behavior.rb
AudreyAltman: I second that :+1:

AudreyAltman:
This looks :rocket: to me.

PR #122, opened 2015-03-17
Implement MDL Harvester ("generic" API harvester)

no-reply:
This might benefit from a quick review from @AudreyAltman or @markbreedlove.

It works fine as an MDL harvester, and I'm thinking a bit about how to separate MDL specific issues from general REST `ApiHarvester` concerns.


no-reply:
I think this is ready for a more serious review, at this point.

markbreedlove:
Since we have decided to put provider-specific mappings into `/vendor/mappings` in the upstream application, what to you think about doing something similar with harvesters?

no-reply:
That's an interesting thought.  At the very least, I think the MDL harvester belongs in `heidrun`.

lib/krikri/harvesters/api_harvester.rb
```diff
@@ -0,0 +1,164 @@
...
+    ##
+    # @return [Hash] A hash documenting the allowable options to pass to
+    #   initializers.
+    #
+    # @see Krikri::Harvester::expected_opts
+    def self.expected_opts
+      {
+        key: :api,
+        opts: {
+          params: { type: :string, required: false }
```
markbreedlove: It looks like `:params` might want to be specified as a Hash, because of the way it's dereferenced in `#next_options`

no-reply: Good catch.

markbreedlove:
The `get_` methods that you added make it easier to subclass this.

The methods of the base class imply that the requests are going to vary by pagination.  I'm trying to imagine how we'd approach something like NYPL, where there is an initial request for a list of sets, and then a request for each set, which gets a long list of record IDs, and then individual requests for the record resources.  As I recall, the NYPL API is pretty much a REST API.  We'd have to override a number of these methods, including `enumerate_records`.  `next_options` would also have to be overridden.

I'm just thinking out loud.  Should there be some notion of a "set" or collection as part of this class?  Or, at at higher level, should this class concern itself with the variation of more parameters than a page or record count?

no-reply:
> I'm just thinking out loud. Should there be some notion of a "set" or collection as part of this class? Or, at at higher level, should this class concern itself with the variation of more parameters than a page or record count?

Probably.  My thinking was to extract this into something that is plausibly reusable... I expect it will change substantially when we come to the next REST API.  Or maybe this will turn out to be less generic than we would hope.

I suppose we'll see?

no-reply:
I'm going to resubmit this with the MDL stuff moved out into `Heidrun`.

markbreedlove:
Related to https://github.com/dpla/KriKri/pull/130

PR #123, opened 2015-03-17
Improve tests and documentation in OAIHarvester

no-reply:
Slightly improves testing for laziness of multi-set harvests. Adds some doc annotations to OAIHarvester methods.

These are just some minor improvements I made while diagnosing another issue.


AudreyAltman:
:+1:

PR #124, opened 2015-03-17
Revert "Add controller/view for Krikri::Activity; add PrettifierHelperBe...

no-reply:
Reverts dpla/KriKri#121


no-reply:
#121 fails with [this error](https://dpla.slack.com/archives/tech/p1426622104000123): 

> At `/krikri/activities` I am getting `Missing partial kaminari/twitter-bootstrap-3/_paginator with {:locale=>[:en], :formats=>[:html], :variants=>[], :handlers=>[:erb, :builder, :raw, :ruby, :coffee, :jbuilder]}.`

We need to
  - diagnose that issue and patch
  - introduce (& document) some kind of view testing

PR #125, opened 2015-03-18
Add support for additional date formats

no-reply:
The date parser now works with variations on 'circa' and some other random date formatting quirks.

### DIFF
--- a/lib/krikri/enrichments/timespan_split.rb
+++ b/lib/krikri/enrichments/timespan_split.rb
@@ -28,18 +28,40 @@ module Krikri::Enrichments
   class TimespanSplit
     include Krikri::FieldEnrichment
 
+    ##
+    # Enrich a `DPLA::MAP::TimeSpan` object or string value with `begin` and
+    # `end` values.
+    #
+    # @param value [DPLA::MAP::TimeSpan, String, Object]
+    #
+    # @return [Object] a new `TimeSpan` object containing the providedLabel
+    #   and the enriched begin/end; if given a value other than a `TimeSpan`
+    #   or `String` returns that value.
     def enrich_value(value)
       value = timespan_from_string(value) if value.is_a? String
       return value unless value.is_a? DPLA::MAP::TimeSpan
       populate_timespan(value)
     end
 
+    ##
+    # Converts a string to a `DPLA::MAP::TimeSpan` with the string as
+    # `providedLabel`.
+    #
+    # @param [String] a string value containing a date, time, or timespan
+    #
+    # @return [DPLA::MAP::TimeSpan] a new, empty timespan with `providedLabel`
     def timespan_from_string(value)
       timespan = DPLA::MAP::TimeSpan.new
       timespan.providedLabel = value
       timespan
     end
 
+    ##
+    # Populates a timespan with a begin and end date.
+    #
+    # @param timespan [DPLA::MAP::TimeSpan]
+    #
+    # @return [DPLA::MAP::TimeSpan]
     def populate_timespan(timespan)
       return timespan unless (timespan.begin.empty? || timespan.end.empty?) &&
         !timespan.providedLabel.empty?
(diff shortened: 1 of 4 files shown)


lib/krikri/util/extended_date_parser.rb
```diff
@@ -21,17 +21,18 @@ module ExtendedDateParser
...
+      date = parse_interval(str) if allow_interval
+      date ||= parse_m_d_y(str)
+      date ||= Date.edtf(str.gsub('.', '-'))
+      date ||= partial_edtf(str)
+      date ||= decade_hyphen(str)
+      date ||= month_year(str)
+      date ||= decade_s(str)
+      date ||= hyphenated_partial_range(str)
+      date ||= parse_date(str)
+      date ||= circa(str) if str.match(/[circabout]/i)
```
markbreedlove: Is this going to match various month names?  If so, is it worth being more specific?

lib/krikri/util/extended_date_parser.rb
```diff
@@ -55,6 +56,22 @@ def range_match(str)
       end
     end
 
+    def preprocess(str)
+      str.gsub!(/late/i, '')
+      str.strip!
+      str.gsub!(/\s+/, ' ')
+      str.gsub!('-', 'x') if str.match(/^[1-9]+\-+$/)
+      str.gsub!('0s', 'x') if str.match(/^[1-9]+0s$/)
```
markbreedlove: Are these characters being turned into "x" so that they can be matched more easily in the other functions?

If so, could comments be added wherever this applies?

lib/krikri/enrichments/timespan_split.rb
```diff
@@ -61,7 +61,10 @@ def parse_labels(labels)
 
     def span_from_date(date)
       return [nil, nil] if date.nil?
-      return [date, date] if date.is_a? Date
+      if date.is_a?(Date)
+        return [date, date] if date.precision == :day
+        return [date, (date.succ - 1)]
```
markbreedlove: I looked up the `Date` documentation to find the definition of `#succ`, which is the succeeding date.  Isn't the succeeding date minus one day the same as `date`?  I could be missing something, but ... my main point: a comment might help to make this clear :-)

markbreedlove:
This obviously adds a lot of useful parsing for these more imprecise dates, which is great.  This is useful and hard to deal with.

It would be nice to have more comments, in general, especially for `#preprocess` and some of the regular expressions.

Is there a dependency in some of the methods upon the translation of certain characters into "x"es?  If so, it would be good to flag that in comments.

lib/krikri/util/extended_date_parser.rb
no-reply: `x` is the EDTF format.

lib/krikri/enrichments/timespan_split.rb
no-reply: `#succ` is overridden by `EDTF` to allow for month/year precision.  So this is either:
  - +1 day, -1 day
  - +1 month, -1 day
  - +1 year, -1 day

For example: If the date is `1999` with year precision, this sets the range to Jan 1, 1999 - Dec. 31 1999.

no-reply:
Thanks for the review.  If you think this is complicated, you should look at BPL's approach!  

I'll make another pass with this to add some comments; they'll come in handy in the long run.  I also have designs on more serious EDTF support (in the form of `RDF::Literal` handlers) that will simplify all of this.

lib/krikri/util/extended_date_parser.rb
no-reply: I'm not sure what this comment refers to, but assuming it's `/[circabout]/i`, it just matches any one of those individual characters.  If nothing else has produced a date match, this method call tries to remove variations on "circa" or "about", reparse the date, and add an EDTF uncertainty flag.

markbreedlove: Right, `/[circabout]/i` matches any one of those characters in the string, such that it would match most month names  or abbreviations.

no-reply: ah, gotcha... month names should already be parsed.  Also, this is just a filter to determine whether to call the method.  It doesn't actually strip anything.

markbreedlove:
Yes, date processing can be really tricky, and this is not compliated.  This PR is great; it's not unclear.

no-reply:
Let me know if these comments still leave something to be desired.

markbreedlove:
It's fine.  I'm realizing that long multiline regex in  `#range_match` was not part of this PR, so I wouldn't advocate changing it here.  Comments could be added to that later at some point, but it's not a big deal.

PR #126, opened 2015-03-18
Feature/punctuation cleanup

no-reply:
Adds several enrichments to cleanup punctuation.

The aim is to be able to conform (at least) to the following:

Titles
--- 
 - Strip out any leading colons or semi-colons. 
 - Strip out any ending punctuation, excluding parentheses but including periods. If we could make some exceptions to the stripping of trailing periods that would be great. For example, we wouldn't strip trailing periods from subject or formats where:
   - A value ends with a space-single letter-period combination ("Smith, John Q." not "Smith, John Q") OR 
   - A value ends with a space-two letters-period combination ("Johnson, Steven, Jr." not "Johnson, Steven Jr"; "oil painting, 64 x 64 in." not "oil painting, 64 x 64 in")
   - All other punctuation within the text should stay as is.

Descriptions
---
 - Strip out any leading colons or semi-colons. 

Subjects, Formats, Creators
---
 - Strip out any leading punctuation except quotation marks.
- Strip out any ending punctuation, excluding parentheses but including periods. If we could make some exceptions to the stripping of trailing periods that would be great. For example, we wouldn't strip trailing periods from subject or formats where:
   - A value ends with a space-single letter-period combination ("Smith, John Q." not "Smith, John Q") OR 
   - A value ends with a space-two letters-period combination ("Johnson, Steven, Jr." not "Johnson, Steven Jr"; "oil painting, 64 x 64 in." not "oil painting, 64 x 64 in")

Date
---
 - Strip out trailing punctuation in this field, except for square brackets, when we're displaying the ProvidedLabel (or whatever it's called).

### DIFF
--- a/lib/krikri/enrichments/strip_ending_punctuation.rb
+++ b/lib/krikri/enrichments/strip_ending_punctuation.rb
@@ -1,0 +1,21 @@
+module Krikri::Enrichments
+  ##
+  # Strip ending punctuation
+  #
+  #   StripEndingPunctuation.new
+  #     .enrich_value("moomin!...!;,.",)
+  #   # => "moomin"
+  #
+  # Leaves quotation marks and closing parentheses & brackets. Also
+  # leaves periods when they follow a one or two letter abbreviation.
+  class StripEndingPunctuation
+    include Krikri::FieldEnrichment
+
+    def enrich_value(value)
+      return value unless value.is_a? String
+      value.gsub!(/[^\p{Alnum}\'\"\)\]\}]*$/, '') unless
+        value.match /\s*[a-z]{1,2}\.$/i
+      value
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


markbreedlove:
Could there be older resources where first names are abbreviated, like "Washington, Geo." for "Washington, George"?  This would also strip "Esq."  Esquire might be obvious without the period, but "Geo" might not be.  There are probably other examples that I can't think of at the moment.

markbreedlove:
I think this is good to go, if nobody thinks that abbreviated proper names need to be fixed, or as soon as that issue can be addressed.

no-reply:
> Could there be older resources where first names are abbreviated, like "Washington, Geo." for "Washington, George"? 

Probably.  I'm basing this on the specifications above, and I imagine we'll catch other punctuation cleanup patterns in the future.  This gets us at least to parity with the old system.

markbreedlove:
That's all fine, then.  I agree that it should just get us to where we were with the old system.

no-reply:
:cool: these are three distinct enrichments in three commits, so merge at your leisure.

PR #127, opened 2015-03-19
Add a genre filter enrichment

no-reply:
Removes values that do not match a given list of genre terms.

Also splits string enrichment shared examples to deal with string enrichments that don't want to ignore non-string values. The shared example which requires that is put into its own group so it can be combined as needed.

### DIFF
--- a/lib/krikri/enrichments/genre_filter.rb
+++ b/lib/krikri/enrichments/genre_filter.rb
@@ -3,7 +3,6 @@ module Krikri::Enrichments
   # Enrichment to remove non-genre fields from
   #
   #   StripHtml.new.enrich_value('Book') => 'Book'
-  #   StripHtml.new.enrich_value('book') => 'book'
   #   StripHtml.new.enrich_value('not a book') => nil
   #
   # Allowed genre terms are:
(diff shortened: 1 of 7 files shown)


lib/krikri/enrichments/genre_filter.rb
```diff
@@ -0,0 +1,46 @@
+module Krikri::Enrichments
+  ##
+  # Enrichment to remove non-genre fields from
+  #
+  #   StripHtml.new.enrich_value('Book') => 'Book'
+  #   StripHtml.new.enrich_value('book') => 'book'
```
markbreedlove: It looks from lines 39-43 like this will result in "Book"  with a capital "B".

no-reply: That's right... this comment doesn't belong, I think.

markbreedlove:
Looks good to go, unless that comment needs to be changed.

no-reply:
I'll change it.

no-reply:
Comment fixed. I think this should stay two commits, so unless there's disagreement there, this is ready to :merge: from my standpoint.

PR #128, opened 2015-03-19
Add Enricher, a SoftwareAgent that runs enrichments

markbreedlove:
One Enricher run chains together multiple enrichments that get run on each record that was generated by a mapping activity.  The enrichments are specified in the options hash that is passed to `Enricher.enqueue`.

### DIFF
--- a/lib/krikri/enricher.rb
+++ b/lib/krikri/enricher.rb
@@ -1,0 +1,158 @@
+module Krikri
+  ##
+  # A SoftwareAgent that runs enrichment processes.
+  #
+  # @example
+  #
+  #   To enrich records that were mapped by the mapping activity with ID 3:
+  #
+  #   # Define which enrichments are run, and thier parameters:
+  #   chain = {
+  #     'Krikri::Enrichments::StripHtml' => {
+  #       input_fields: [{sourceResource: :title}]
+  #     },
+  #     'Krikri::Enrichments::StripWhitespace' => {
+  #       input_fields: [{sourceResource: :title}]
+  #     }
+  #   }
+  #   Krikri::Enricher.enqueue({
+  #     generator_uri: 'http://ldp.local.dp.la/ldp/activity/3',
+  #     chain: chain
+  #   })
+  #
+  # @see Krikri::SoftwareAgent#enqueue
+  # @see Krikri::Enrichment
+  #
+  class Enricher
+    include SoftwareAgent
+
+    attr_reader :chain, :generator_uri
+
+    def self.queue_name
+      :enrichment
+    end
+
+    ##
+    # Create a new Enricher, given a hash of options:
+    #   generator_uri:  the LDP URI of the Activity that generated the mapped
+    #      records that this one will enrich.
+    #   chain:  a hash specifying the input_fields and output_fields, as
+    #      illustrated above, which will be passed to the Enrichment.
+    #
+    # @see Krikri::Enrichment
+    # @param opts [Hash] a hash of options
+    def initialize(opts = {})
+      @generator_uri = RDF::URI(opts.fetch(:generator_uri))
+      @chain = deep_sym(opts.fetch(:chain) { {} })
+    end
+
+    ##
+    # Run the enrichmnt.
+    #
+    # Take each record that was affected by the activity defined by our
+    # instantiation, and apply each enrichment from the enrichment chain.
+    #
+    def run(activity_uri = nil)
+      log :info, 'enricher is running'
+      # see TODO below
+      target_aggregations.each do |agg|
+        begin
+          chain_enrichments!(agg)
+          agg << RDF::Statement(agg, RDF::PROV.wasGeneratedBy, activity_uri) if
+            activity_uri
+          agg.save
+        rescue => e
+          log :error, "Enrichment error: #{e.message}\n#{e.backtrace}"
+        end
+      end
+      log :info, 'enricher is done'
+    end
+
+    # TODO:  remove this when the current topic branch that introduces the
+    # EntityConsumer mixin has been merged.
+    def target_aggregations
+      query = Krikri::ProvenanceQueryClient.find_by_activity(generator_uri)
+      query.execute.lazy.flat_map do |solution|
+        agg = DPLA::MAP::Aggregation.new(solution.record.to_s)
+        agg.get
+        agg
+      end
+    end
+
+    ##
+    # Given an aggregation, take each enrichment specified by the `chain'
+    # given in our instantiation, and apply that enrichment, with the given
+    # options, modifying the aggregation in-place.
+    #
+    def chain_enrichments!(agg)
+      chain.keys.each do |e|
+        enrichment = e.to_s.constantize.new
+        if enrichment.is_a? Krikri::FieldEnrichment
+          agg = do_field_enrichment(agg, enrichment, chain[e])
+        else
+          agg = do_basic_enrichment(agg, enrichment, chain[e])
+        end
+      end
+    end
+
+    private
+
+    ##
+    # Perform a default enrichment, using Enrichment#enrichment or a derived
+    # class that expects the same arguments.
+    #
+    # @param agg [DPLA::MAP::Aggregation]
+    # @param enrichment [Krikri::Enrichment]
+    # @param options [Hash]
+    #
+    # @see Krikri::Enrichment
+    #
+    def do_basic_enrichment(agg, enrichment, options)
+      enrichment.enrich(
+        agg, options[:input_fields], options[:output_fields]
+      )
+    end
+
+    ##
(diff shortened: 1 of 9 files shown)


lib/krikri/enricher.rb
```diff
@@ -0,0 +1,132 @@
...
+    #                        arguments list
+    #
+    def do_field_enrichment(agg, enrichment, options)
+      enrichment.enrich(agg, *options[:input_fields])
+    end
+
+    ##
+    # Transform the given hash recursively by turning all of its string keys
+    # and values into symbols.
+    #
```
markbreedlove: I debated putting this in `lib/util` but I'm not sure that it would be generally useful.

no-reply:
I'm trying this out with ESDN data, now.

spec/lib/krikri/enricher_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+  end
+
+  describe '#initialize' do
+    context 'with a chain argument that has been parsed from JSON' do
+      # Activity records will contain the serialized chain hash, which is
+      # passed in the call to Krikri::Enricher.enqueue.
+      include_context 'common'
+      let(:json_chain) do
+        s = <<-EOS.gsub /\s+/, ' '
```
no-reply: Why the variable? Wouldn't this be just as good?

```ruby
let(:json_chain) do
  <<-EOS.gsub /\s+/, ' '
...
EOS
end
```

spec/lib/krikri/enricher_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  describe '#chain_enrichments!' do
+    context 'with a chain of enrichments' do
+      include_context 'common'
+      let(:aggregation) { build(:aggregation) }
+      subject do
+        described_class.new generator_uri: mapping_activity_uri,
+                            chain: chain
+      end
+      before(:each) do
```
no-reply: `:each` isn't needed; it's implied by `before do`.

lib/krikri/enricher.rb
```diff
@@ -0,0 +1,132 @@
...
+      @generator_uri = RDF::URI(opts.fetch(:generator_uri))
+      @chain = deep_sym(opts.fetch(:chain) { {} })
+    end
+
+    def run(activity_uri = nil)
+      log :info, 'enricher is running'
+      # see TODO below
+      target_aggregations.each do |agg|
+        chain_enrichments!(agg)
+        agg << RDF::Statement(agg, RDF::PROV.wasGeneratedBy, activity_uri) if
```
no-reply: We need a different predicate, here.  See, http://www.w3.org/TR/prov-o/#wasGeneratedBy

> Generation is the completion of production of a new entity by an activity. This entity did not exist before generation and becomes available for usage after this generation.

I'm looking into what should replace this.

spec/lib/krikri/enricher_spec.rb
markbreedlove: Yes.  I'm not sure what the context was, but I recall playing around with heredocs in `irb` and getting some kind of error that went away when I used an assignment.  I'll change this.

lib/krikri/enricher.rb
```diff
@@ -0,0 +1,132 @@
...
+          activity_uri
+        agg.save
```
no-reply: I'm not sure if this is a blocker, but it may be desirable to only save if something has actually changed.  Rerunning enrichments is currently creating new versions even when the graph is totally isomorphic.

no-reply:
This is working more or less perfectly.  I've added a few comments, and have a lot of wheels turning as I play with this.  More comments on the way, but I think we can target this for merge by the end of tomorrow so we can run more serious tests over the weekend.

lib/krikri/enricher.rb
markbreedlove: Yes, that would be great.  Is there an easy way to tell if something has changed?  Checksum the aggregation before and after?

lib/krikri/enricher.rb
```diff
@@ -0,0 +1,132 @@
...
+      end
+    rescue => e
+      Rails.logger.error("Enrichment error: #{e.message}\n#{e.backtrace}")
```
no-reply: Have you thought about error handling like happens in the `Harvesters` and `Mapper::Agents`?  I had a random marmotta 500 error that killed a run, and it might be better just to pass on?

markbreedlove: Yes, it would be best to continue on, like `Harvester` does.  When I was looking at `Mapper::Agent#run`, I probably overlooked the fact that the rescue is inside of the loop in that method.  I'll change this to catch the exception inside of the `each` proc.

lib/krikri/enricher.rb
no-reply: Yeah, I'm surprised to not be able to think of a better way to do this and/or a library that can handle it.  I wonder whether there's something for this in `ActiveSupport`?  

If not, I'm okay with leaving it here as a private method for now.  The main alternative seems to be to reopen `Hash` and convert this to `#deep_map(&block)`.

markbreedlove: My main thing is simply that I have a nagging feeling that I may be overlooking something that code outside of this context would assume.  It's fine for this class, where the cases are very limited, but I don't know about encouraging its use elsewhere.

no-reply:
599151a deals with https://github.com/dpla/KriKri/pull/128#discussion_r26784758 and adds some tests.

If @markbreedlove or @anarchivist reviews that last commit, I think we're in business on this. :squash: & :merge:.  Probably leave the first and last commits separate when squashing, otherwise use your own discretion.

no-reply:
I also rebased and force pushed.  Sorry @markbreedlove.

markbreedlove:
No problem with the force push.  I'll squash it and then will defer to someone else to merge.  I'll give @anarchivist a chance to review it.

lib/krikri/enricher.rb
```diff
@@ -0,0 +1,131 @@
...
+  class Enricher
+    include SoftwareAgent
+
+    attr_reader :chain, :generator_uri
+
+    def self.queue_name
+      :enrichment
+    end
+
+    def initialize(opts = {})
```
anarchivist: Documentation?

lib/krikri/enricher.rb
```diff
@@ -0,0 +1,131 @@
...
+      @generator_uri = RDF::URI(opts.fetch(:generator_uri))
+      @chain = deep_sym(opts.fetch(:chain) { {} })
+    end
+
+    def run(activity_uri = nil)
```
anarchivist: Documentation?

lib/krikri/enricher.rb
```diff
@@ -0,0 +1,131 @@
...
+    def target_aggregations
+      query = Krikri::ProvenanceQueryClient.find_by_activity(generator_uri)
+      query.execute.lazy.flat_map do |solution|
+        agg = DPLA::MAP::Aggregation.new(solution.record.to_s)
+        agg.get
+        agg
+      end
+    end
+
+    def chain_enrichments!(agg)
```
anarchivist: Documentation?

lib/krikri/enricher.rb
```diff
@@ -0,0 +1,131 @@
...
+    private
+
+    ##
+    # Perform a default enrichment, using Enrichment#enrichment or a derived
+    # class that expects the same arguments.
+    #
+    # @param agg [DPLA::MAP::Aggregation]
+    # @param enrichment [Krikri::Enrichment]
+    # @param options [Hash]
+    #
```
anarchivist: perhaps a `@see` reference to `Krikri::Enrichment` here?

lib/krikri/enricher.rb
anarchivist: Would [`Hash#deep_symbolize_keys`](http://api.rubyonrails.org/classes/Hash.html#method-i-deep_symbolize_keys) from `ActiveSupport`'s core extensions for `Hash` work?

markbreedlove: No, because it only symbolizes keys, not the values, as well.

no-reply:
I'm :+1: on this, if @anarchivist's comment about `TODO` is resolved (I'm not clear on what's being asked), then it's ready once the docs commit is :squash:ed.

anarchivist:
Yeah, I'm good. :ok:

no-reply:
This is now :squash:'d.  I'm inclined to let @anarchivist do the :merge:, since this has significant work from both @markbreedlove and myself, and he's already familiar. :)

anarchivist:
:money_with_wings: :rocket:

no-reply:
Awesome work! This code is a homerun.

markbreedlove:
:-)

PR #130, opened 2015-03-23
Implement a "generic" API harvester

no-reply:
This implements a base harvester for REST APIs, targeted initially at serving the MDL API. Some attention has been given to making it easy to subclass `ApiHarvester` and customize it to a given API, but more work will probably be needed on the harvester and the relevant specs when other implementations are written.


markbreedlove:
Related to https://github.com/dpla/KriKri/pull/122

Following up on the conversation there:  I agree that we should wait and see what we need to do surrounding sets or collections.  The current class assumes pagination as the only variation from one request to another, but I think we should start with this, especially so that we can get moving with MDL.

no-reply:
:+1: I think the only alternative is to try to write another API harvester now, and that seems like it can wait.

PR #131, opened 2015-03-23
Add and refactor search indexing; add behaviors

markbreedlove:
- Add behaviors (in `entity_behaviors`) to provide different answers to generated-entities questions depending on the type of activity.
- Add `EntityConsumer` mixin for classes, like certain software agents, that generate entities.
- Update `Krikri::Mapper::Agent` because of the changes above
- Replace `IndexService` with `SearchIndex` and its subclasses, `QASearchIndex` and `ProdSearchIndex`
- Add `Activity#generated_entitiy_uris`, which enumerates URI strings of entities that pertain to the activity.
- Add `Activity#generated_entities`, which returns entities (for example, aggregations) that were affected by the activity.  A behavior's `#generated_entities` method is invoked for this purpose.
- Add the ElasticSearch service to the development VM.  Add tests that are not runnable in Travis CI, but that work in the VM with ElasticSearch 0.90 (which can't be installed in a Travis container).

The buildout of search indexing has required us to think about how we can treat provenance consistently for different kinds of activities.  The new behavior classes and the additions of the Activity model make it possible to ask for the entities that were affected by an activity, using consistent interfaces.  A mapping activity may have consumed the original records generated by a harvest, but an enrichment and an indexing activity are usually going to consume aggregations, when KriKri is used within the DPLA context.  In the case of another organization, it's possible that the entities generated by mappings and enrichments could be something else, thus the level of abstraction that you see here.

## Questions / Issues:

- Is the naming of `QASearchIndex` and `ProdSearchIndex` appropriate, considering how the architecture could be changed in the future?  What if we change to using ElasticSearch or Solr for both the QA and production indices?
- I have a [doc comment for `EntityConsumer#set_generator_activity!`](https://github.com/dpla/KriKri/blob/feature/index-an-activity/lib/krikri/entity_consumer.rb#L25-L27) that I added in an early commit, but forgot to remove when I removed the functionality it refers to: the handling of a `generator_activity` option.  I was just about to clean it up when I realized why I had that functionality there originally. An indexing activity is not going to create a provenance record in Marmotta.  Therefore, if there's any `EntityConsumer` that wants to act upon it, the `generator_uri` parameter that is normally passed in the options hash to an `EntityConsumer` will not be useful.  It will be necessary to pass `#set_generator_activity!`  an activity object or id, not a `generator_uri`.  I think that would be more efficient than saving additional provenance data in Marmotta when the record is indexed.  It's possible that we should come back to this later, if we ever have a consumer that wants to feed on _indexed_ records.
- Related to the point above, `EntityConsumer` has the option `generator_uri`, though it might be renamed to something else, because entities can be _revised_ by, in addition to generated by; and possibly more.


markbreedlove:
I forgot to mention this, but [`Krikri::ProdSearchIndex#hash_for_index_schema`](https://github.com/dpla/KriKri/blob/feature/index-an-activity/lib/krikri/search_index.rb#L306-L313) will need to be updated as part of [our upcoming story to do the MAPv4 to MAPv3.1 crosswalk](https://www.pivotaltracker.com/story/show/84399108).  Before that happens, tests will work because they use a factory that spits out MAP3.1-compliant JSON, but an indexing run for a development or staging environment shouldn't work.

app/models/krikri/activity.rb
```diff
@@ -83,5 +83,23 @@ def rdf_subject
       RDF::URI(Krikri::Settings['marmotta']['ldp']) /
         Krikri::Settings['prov']['activity'] / id.to_s
     end
+
+    ##
+    # Return an Enumerator of URI strings of entities (e.g. aggregations or
+    # original records) that pertain to this activity
+    #
+    # @return [Enumerator] URI strings
+    def generated_entity_uris
```
no-reply: We are now working with both "generated" and "revised" entities.  I'm not sure whether this merits a change in these method signatures, but I thought it was worth mentioning.

I'd say we either keep these as is (with the excuse that PROV says that "Activities generate Entities"), or change them to just `#entity_uris` and `#entities`.  Thoughts?

app/models/krikri/activity.rb
```diff
@@ -83,5 +83,23 @@ def rdf_subject
...
+      activity_uri = RDF::URI(rdf_subject)  # This activity's LDP URI
+      query = Krikri::ProvenanceQueryClient.find_by_activity(activity_uri)
+      query.each_solution.lazy.map do |s|
+        s.record.to_s
+      end
+    end
+
+    def generated_entities
```
no-reply: Doc?

lib/krikri/enricher.rb
```diff
@@ -33,6 +34,13 @@ def self.queue_name
     end
 
     ##
+    # @see Krikri::Activity#generated_entities
+    # @see Krikri::EntityBehavior
+    def entity_behavior
+      Krikri::AggregationEntityBehavior
```
no-reply: Is there a way around hard coding this?  Maybe something like

```ruby
attr_writer :entity_behavior

def entity_behavior
  @entity_behavior ||= Krikri::AggregationEntityBehavior
end
```

This isn't hugely important, since the Enrichments in their current form only work on `ActiveTriples::Resources`; but it would be nice to loosen this dependency.

app/models/krikri/activity.rb
```diff
@@ -83,5 +83,23 @@ def rdf_subject
...
+    ##
+    # Return an Enumerator of URI strings of entities (e.g. aggregations or
+    # original records) that pertain to this activity
+    #
+    # @return [Enumerator] URI strings
+    def generated_entity_uris
+      activity_uri = RDF::URI(rdf_subject)  # This activity's LDP URI
+      query = Krikri::ProvenanceQueryClient.find_by_activity(activity_uri)
+      query.each_solution.lazy.map do |s|
+        s.record.to_s
```
no-reply: What's the benefit of the `#to_s`? Would it be more flexible to leave these as `RDF::URI`s?

lib/krikri/enrichment.rb
```diff
@@ -3,6 +3,7 @@ module Krikri
   # Mixin module for enriching a set of input_fields and setting the resulting
   # values to a set of output fields.
   module Enrichment
+
```
no-reply: maybe just reset this file; whitespace only diff.

app/models/krikri/activity.rb
markbreedlove: I think renaming this would be best, because "generated" could definitely be confusing for someone new to the code, if there are expectations set elsewhere for a distinction between "revised" and "generated."  There aren't too many lines that need changing.

app/models/krikri/activity.rb
markbreedlove: Thanks; will add that ...

lib/krikri/entity_behavior.rb
```diff
@@ -0,0 +1,18 @@
+
+module Krikri
+  ##
+  # Base class for behaviors related to entities that are generated by
+  # activities.
+  class EntityBehavior
```
no-reply: This could do with method/attribute documentation, and probably an `@example` in the class level doc.

Consider using: http://www.rubydoc.info/gems/yard/file/docs/Tags.md#abstract

lib/krikri/entity_consumer.rb
```diff
@@ -0,0 +1,50 @@
...
+        # allow generator_uri to be string or RDF::URI with `to_s' ...
+        activity_id = generator_uri.to_s[/\d+$/].to_i  # 0 if no match
+        fail "Can not determine ID for #{generator_uri}" if activity_id == 0
+        @generator_activity = Krikri::Activity.find_by_id(activity_id)
+        raise "Generator activity not found for id #{activity_id}" \
+          if !@generator_activity
+      end
+    end
+
+    def generator_activity
```
no-reply: `attr_reader :generator_activity` instead?

lib/krikri/mapper.rb
```diff
@@ -90,23 +90,51 @@ def mapping_exception_desc(rec)
     ##
     # A SoftwareAgent to run mapping processes.
     #
+    # @example
```
no-reply: Nice!

lib/krikri/mapper.rb
```diff
@@ -90,23 +90,51 @@ def mapping_exception_desc(rec)
...
 
       def self.queue_name
         :mapping
       end
 
+      ##
+      # @see Krikri::Activity#generated_entities
+      # @see Krikri::EntityBehavior
+      def entity_behavior
+        Krikri::AggregationEntityBehavior
```
no-reply: see comment about hard coding, here: https://github.com/dpla/KriKri/pull/131/files#r27034567

spec/lib/krikri/mapper_agent_spec.rb
```diff
@@ -69,25 +109,4 @@
     end
   end
 
-  # TODO: these tests are tied closely to implementation.
```
no-reply: Hey! You got rid of my bad tests! 

Thanks!

no-reply:
> I have a doc comment for EntityConsumer#set_generator_activity! that I added in an early commit, but forgot to remove when I removed the functionality it refers to: the handling of a generator_activity option. I was just about to clean it up when I realized why I had that functionality there originally. An indexing activity is not going to create a provenance record in Marmotta. Therefore, if there's any EntityConsumer that wants to act upon it, the generator_uri parameter that is normally passed in the options hash to an EntityConsumer will not be useful. It will be necessary to pass #set_generator_activity! an activity object or id, not a generator_uri. I think that would be more efficient than saving additional provenance data in Marmotta when the record is indexed. It's possible that we should come back to this later, if we ever have a consumer that wants to feed on indexed records.

This all sounds right.  I think its safe to assume for the moment indexing is not an activity we track. It doesn't generate/revise anything (except actual index documents, which I can't see any reason to have URIs for).  Let's say this is not a valid provenance query, for now.  We can revisit if something changes.

no-reply:
This is great.  The main thing I see at this point is to document `EntityBehavior` and its implementations.

Coveralls also marks this as -0.8% test coverage on the project, which I'd like to fix.  I'll take another pass to look at tests in a bit.

markbreedlove:
Re. the doc comment for `EntityConsumer#set_generator_activity!`, I agree on coming back to it later, and I'll revise the doc comment to indicate it's something that might become an issue in the future.

lib/krikri/entity_consumer.rb
markbreedlove: Oh, can you do that in a module?  (`EntityConsumer` is not a class.)

lib/krikri/entity_behaviors/aggregation_entity_behavior.rb
```diff
@@ -0,0 +1,12 @@
+
+module Krikri
+  class AggregationEntityBehavior < Krikri::EntityBehavior
+    def generated_entities
+      @activity.generated_entity_uris.lazy.map do |uri|
+        agg = DPLA::MAP::Aggregation.new(uri)
+        agg.get   # slow?
```
no-reply: I think this is significantly less expensive than `OriginalRecord#load`, but still possibly something to be avoided here.  This call is already lazy, so the question is whether you'll ever want to access `#generated_entities` and get back unloaded objects (i.e. objects of the correct type, with the correct URI, but unpopulated).  I suspect the answer may be yes, but I don't see any use cases immediately.

Maybe the answer is something like:

```ruby
def entities(load = true)
  @activity.generated_entity_uris.lazy.map do |uri|
    agg = DPLA::MAP::Aggregation.new(uri)
    agg.get  if load
    agg
  end
end
```

Similarly for `OriginalRecord` you could do `load ? OriginalRecord.load(uri) : OriginalRecord.new(uri)`.

lib/krikri/entity_consumer.rb
no-reply: I'm not sure of the top of my head... but if not, something like this should work:

```ruby
included { attr_reader :generator_activity }
```

spec/lib/krikri/mapper_agent_spec.rb
markbreedlove: :zap:

app/models/krikri/activity.rb
markbreedlove: `Krikri::OriginalRecord#initialize` wants to work with a string, or something that has `#to_s`.  I must have figured I'd cast it as a string if that's the way it was going to be used.   `ActiveTriples::Resource#set_subject!` takes `uri_or_str` too, so think that I decided to cast it as a string at this point to make sure it _can_ be cast as a string, rather than letting any error with that result in a longer stacktrace.

I don't have anything against letting it be an `RDF::URI`, it that would be useful in the future, but right now all it's being used for is instantiating Aggregations or OriginalRecords.

lib/krikri/entity_behaviors/aggregation_entity_behavior.rb
markbreedlove: Thanks for that idea.  I'm pushing a commit that will add this `load` argument to both methods.

spec/lib/krikri/aggregation_entity_behavior_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    # Krikri::Mapper::Agent is instantiated, but that seems like jumping
+    # through more hoops and making this less comprehensible than it is now.
+    DatabaseCleaner.clean_with(:truncation)
+    create(:krikri_harvest_activity)
+  end
+
+  describe '#entities' do
+
+    let(:mapping_activity_uri) do
+      (RDF::URI(Krikri::Settings['marmotta']['ldp']) /
```
anarchivist: This pattern that creates the activity URIs feels somewhat brittle. Do the specs themselves necessarily need to have awareness of Krikri's settings to construct activity URIs for testing? Can't we assume that (in this case) that `mapping_activity_uri` would return the same thing as `activity.rdf_subject`?

no-reply: I don't actually see any reason why you couldn't just use a dummy URI, here.  All interactions that rely on the URI are (should/be) stubbed.

`RDF::URI('http://example.org/moomin)` or `double('uri')` seem fine.  This variable also looks badly named--it seems to be a URI for a mapped record, rather than an activity.

anarchivist: Ah, yep. I see what you're saying, based on what it happens in the `it 'enumerates generated entities'` block.

spec/lib/krikri/entity_consumer_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  before(:all) do
+    DatabaseCleaner.clean_with(:truncation)
+    create(:krikri_harvest_activity)
+    create(:krikri_mapping_activity)
+  end
+
+  context 'as a mapper agent' do
+    let(:generator_uri) do
+      (RDF::URI(Krikri::Settings['marmotta']['ldp']) /
```
anarchivist: Same comment as for [`aggregation_entity_behavior_spec.rb`](https://github.com/dpla/KriKri/pull/131/files#r27392504), without @no-reply's feedback.

spec/lib/krikri/indexer_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'spec_helper'
+
+describe Krikri::Indexer do
+  before(:all) do
+    DatabaseCleaner.clean_with(:truncation)
+    create(:krikri_enrichment_activity)
+  end
+
+  enrichment_gen_uri_str = \
+    (RDF::URI(Krikri::Settings['marmotta']['ldp']) /
```
anarchivist: Same comment as for [`aggregation_entity_behavior_spec.rb`](https://github.com/dpla/KriKri/pull/131/files#r27392504), without @no-reply's feedback.

spec/lib/krikri/mapper_agent_spec.rb
```diff
@@ -1,11 +1,50 @@
...
 describe Krikri::Mapper::Agent do
-  opts = { name: :agent_map, generator_uri: 'http://example.org/agent/gen/1' }
-  it_behaves_like 'a software agent', opts
+  before(:all) do
+    DatabaseCleaner.clean_with(:truncation)
+    create(:krikri_mapping_activity)
+    create(:krikri_harvest_activity)
+  end
+
+  harvest_gen_uri_str = \
```
anarchivist: Same comment as for [`aggregation_entity_behavior_spec.rb`](https://github.com/dpla/KriKri/pull/131/files#r27392504) for bot here and `mapping_gen_uri_str`, without @no-reply's feedback.

spec/models/activity_spec.rb
```diff
@@ -147,4 +147,26 @@
...
+
+  describe '#entity_uris' do
+    include_context 'provenance queries'
+    include_context 'entities query'
+    # See spec/factories/krikri_activities.rb
+    # generator_uri is the URI of the harvest activity
+    # generator_uri matches what Krikri::Activity will construct as the
+    # uri, given its value of #rdf_subject, in #aggregations_as_json
+    # See 'provenance queries' shared context.  
+    let(:generator_uri) do
```
anarchivist: Same comment as for [`aggregation_entity_behavior_spec.rb`](https://github.com/dpla/KriKri/pull/131/files#r27392504), without @no-reply's feedback. I wonder if `krikri_activity.rdf_subject` will work here.

anarchivist:
I'm :+1: on this. My comments on the specs may be nitpicky but they will certainly make them easier to manage. @no-reply (in @markbreedlove's absence today) will make the suggested changes, and then rebase and squash as appropriate. Then we can :merge:!

spec/lib/krikri/aggregation_entity_behavior_spec.rb
markbreedlove: I can't really explain this, except to say that I found working with RSpec and mocking up all of the necessary function calls to be very confusing, at least using what was in place before this feature branch was started.  (And I started out simply trying to emulate what was there originally.)  I found that I had to dig quite a lot into the internals of methods I wasn't testing, and, instead of abstracting the process and making it easier to read, a lot of the mocks made it more difficult and time-consuming.  I probably copied and pasted something from somewhere else and was trying to shoehorn it in to fit within the context of this test.  I'll have another look at this as soon as I can today.

markbreedlove: Regarding the dummy URI, `#rdf_subject` responds with a URI that includes the hostname of whatever host is running the tests, so I was trying to make the tests as portable as possible, and I'm sure I didn't realize about the use of `double` at the time.

markbreedlove:
Hi, @anarchivist and @no-reply. How do the last two commits look?

anarchivist:
@markbreedlove :+1:

markbreedlove:
OK, I will squash ...

PR #132, opened 2015-03-24
Add a test that covers Enricher#do_basic_enrichment

markbreedlove:
This method wasn't being covered by unit tests because the only
Enrichments that we have to date are FieldEnrichments.  A new Enrichment
class is mocked-up for the time being.

### DIFF
--- a/spec/lib/krikri/enricher_spec.rb
+++ b/spec/lib/krikri/enricher_spec.rb
@@ -126,5 +126,36 @@ EOS
           .to eq(['nice and clean'])
       end
     end
+
+    context 'with a basic enrichment' do
+      let(:chain) do
+        {
+          'Krikri::Enrichments::BasicEnrichment' => {
+            input_fields: [{sourceResource: {creator: :providedLabel}}],
+            output_fields: [{sourceResource: :creator}]
+          }
+        }
+      end
+      let(:aggregation) { build(:aggregation) }
+
+      subject do
+        described_class.new generator_uri: mapping_activity_uri, chain: chain
+      end
+
+      before do
+        module Krikri::Enrichments
+          class BasicEnrichment
+            include Krikri::Enrichment
+            def enrich_value(value)
+              value
+            end
+          end
+        end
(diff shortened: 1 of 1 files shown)


spec/lib/krikri/enricher_spec.rb
```diff
@@ -126,5 +126,38 @@
           .to eq(['nice and clean'])
       end
     end
+
+    context 'with a basic enrichment' do
+      # TODO:  Revise this when an enrichment is added that directly mixes in
```
no-reply: I think this is great as is. No need to revise.

spec/lib/krikri/enricher_spec.rb
```diff
@@ -126,5 +126,38 @@
...
+      let(:chain) do
+        {
+          # ... When that is done, this class will be changed:
+          'Krikri::Enrichments::BasicEnrichment' => {
+            input_fields: [{sourceResource: {creator: :providedLabel}}],
+            output_fields: [{sourceResource: :creator}]
+          }
+        }
+      end
+      let(:aggregation) { build(:aggregation) }
```
no-reply: Maybe some friendly whitespace lines somewhere in here?

no-reply:
:+1:

Let me know if you don't want to remove the `TODO` and/or disagree about whitespace.  I'm happy to merge this as is if you think it's ready.

markbreedlove:
I'm fine removing the TODOs.  Now that I think about it, it's probably better in theory to mock up a class than to use one particular class as a representative.

As for whitespace, I'm actually prefer more whitespace than what I've been using, but I get rubocop complaints sometimes.  One thing that really bugs me is not having whitespace before the closing `end` token of a block, which seems to be the standard. But I do it anyway because it seems to be what's expected.  The readability, to my eye, depends on the size of the method or module.  I'll add some more space to break out the `before` and `subject`

markbreedlove:
@tom, just force pushed this commit with the changes.

PR #133, opened 2015-04-01
Fixes rake sample record

no-reply:
The arguments for `#add` changed, but this client did not. This fixes that.

I've created a chore for testing our rake tasks to prevent future regressions: https://www.pivotaltracker.com/story/show/91625412


markbreedlove:
OK, that looks good, and thanks for creating the chore!

Is everyone else OK with this?  I can merge it, if so.

AudreyAltman:
I tested this locally, and seems to be saving records to solr correctly.  :rocket:

PR #134, opened 2015-04-01
Feature/qa

no-reply:
This introduces QA reports.

@AudreyAltman (can you give this a look and maybe write a commit comment for 26a9d88 ?)


no-reply:
This supecedes and closes #100.

AudreyAltman:
I am getting an error when I try to load any page in my browser window:  
```query not supported Processing query SELECT * WHERE { _:b3 ?p ?o . }```

I have traced the source of this error to `Krikri::Provider.all`.  When I run it in my console, I get:
```Krikri::Provider.all
Solr query: {:qt=>"search", :rows=>0, :id=>"*:*", "facet.field"=>"provider_id"}
Solr fetch (3.5ms)
SPARQL::Client::ServerError: query not supported Processing query SELECT * WHERE { _:b3 ?p ?o . }
from /Users/audreyaltman/.rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/gems/sparql-client-1.1.4/lib/sparql/client.rb:344:in `block in response'```

**Comment by AudreyAltman** [comment]
@no-reply can you add documentation, to `README` and/or `app/models/krikri/qa_report.rb` explaining how to generate QA reports?

File: app/controllers/krikri/providers_controller.rb [hunk 1]
```diff
@@ -0,0 +1,19 @@
...
+  # This models a provider that is stored in Marmotta.
+  class ProvidersController < ApplicationController
+    def index
+      @providers = Krikri::Provider.all
+    end
+
+    def show
+      if params[:set_session]
+        session[:current_provider] = params[:id]
+        redirect_to :back, provider: params[:id]
```
no-reply: This isn't behaving as expected.  I'm aiming to get it to redirect to the calling page, but with the added provider param.  Perhaps there's a better way of approaching this?

app/controllers/krikri/providers_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+      elsif params[:clear_session]
+        session.delete :current_provider
+        redirect_to providers_path
```
no-reply: This *is* working as expected, but isn't really the desired behavior.  It redirects session variable clears (clicks on `All Providers`) to the providers dashboard, rather than the provider-less version of the calling page.

AudreyAltman:
Provider context doesn't seem to be working on `validation_reports`.  In my browser, with the sample data from the rake task, when I load `http://localhost:3000/krikri/validation_reports/sourceResource_title`, I see one record, as expected :+1: However, when I load `http://localhost:3000/krikri/validation_reports/sourceResource_title?provider=123`, I do not see any records, which is incorrect.  I should see one record.

AudreyAltman:
A single record is not rendering from it's local name.  In my browser, with the sample data from the rake tasks, when I load `http://localhost:3000/krikri/records/http:%2F%2Flocalhost:8983%2Fmarmotta%2Fldp%2Fitems%2Fkrikri_sample_invalid`, I get the expected record :+1:, but we've already learned that URI id's cause breakages if they include dots.  When I do `http://localhost:3000/krikri/records/krikri_sample_invalid` (using the record's local name instead of the URI), I get a `record not found` error.

AudreyAltman:
Provider context does not seem to be working on the the `records#show` page.  In my browser, with the sample data from the rake tasks, when I load `http://localhost:3000/krikri/records/http:%2F%2Flocalhost:8983%2Fmarmotta%2Fldp%2Fitems%2Fkrikri_sample?provider=123`, the random record button reads "Random record from All Providers" instead of "Random record from Moomin valley Historical Society", as I would expect.  Also, the random record button links to any record, not just records with a `provider_id` of `123`.  Finally, after the button is clicked, the new record URI does not contain the param `provider=123` as I would expect.

app/views/krikri/shared/_navigation.html.erb
```diff
@@ -1,10 +1,45 @@
...
+          </a>
+          <ul class='dropdown-menu' role='menu'>
+	    <li>
+              <%= remove_session_provider %>
+            </li>
+            <% if session[:current_provider].present? %>
+              <li>
+                <%= set_session_provider(session[:current_provider]) %>
+              </li>
+              <li class='divider'></li>
```
AudreyAltman: You can delete `<li class='divider'></li>`.  It's original purpose was to separate "All Providers" from individual providers, and it is no longer serving that purpose.

no-reply:
The local name issue (https://github.com/dpla/KriKri/pull/134#issuecomment-88648192) should be fixed by 82cf8d6.  It was originally removed in 548c381d5ee.

AudreyAltman:
Small unexpected behavior:  When I load `http://localhost:3000/krikri/reports?provider=123` into my browser window, I see an error message under the header "QA Reports".  The error message reads, "There was an error generating the QA reports."  I expected the space under the header "QA Reports" to be blank, as it is on `http://localhost:3000/krikri/reports`.

no-reply:
I believe the latest commits deal with all the issues raised by @AudreyAltman's comments.

@AudreyAltman is going to take a look at moving session handling to its own controller, to deal with: https://github.com/dpla/KriKri/pull/134#discussion_r27616482

In the meanwhile, I'm going to try to diagnose the intermittent test failures (I suspect these are related to index clearing) and add documentation throughout.

@anarchivist your review would be very helpful, at this point.

app/controllers/krikri/providers_controller.rb
anarchivist: IIRC, `redirect_to :back` wasn't working here. What about something like `redirect_to(request.referer || providers_path)`?

no-reply: What we're really shooting for is something like

```pseudocode
  redirect_to "#{request.referrer}?provider=#{params[:id]}"
```
 unless `request.referrer` doesn't accept a `provider`.  @AudreyAltman made the strong point that this really needs to be a `POST` request, and i'm of the opinion that `session` should have its own controller.  The POST + redirect could be handled, ideally, in javascript, with only the navbar reloaded if the redirect turns out to be unnecessary.

Anyway, I'm increasingly of the opinion that this should wait.  We can punt it to a story like: 

> As a QA user, I want to be able to select a provider (or "All Providers") from the nav bar, and get the current content filtered to provider.

anarchivist: :+1:

app/models/krikri/provider.rb
```diff
@@ -0,0 +1,80 @@
+module Krikri
+  ##
+  # This models provider data from the {Krikri::Repository}.
+  #
+  # @todo create an appropriate `ActiveTriples::PersistenceStrategy' for
+  #   handling providers as data from a repository alongside LDP data.
+  # @see ActiveTriples::Resource
+  class Provider < DPLA::MAP::Agent
+    configure :base_uri => 'http://dp.la/api/contributor/'
```
anarchivist: This should probably be configurable if it's going to sit in Krikri, correct?

app/models/krikri/provider.rb
```diff
@@ -0,0 +1,80 @@
...
+    end
+
+    ##
+    # A list of records of type ore:Aggregation associated with this provider
+    #
+    # @return [Array<DPLA::MAP::Aggregation>] the Aggregations with this
+    #   provider
+    def records
+      query = Krikri::Repository.query_client.select
+              .where([:record, RDF::EDM.provider, self],
```
anarchivist: wowwwww very cool!

app/models/krikri/provider.rb
no-reply: Yes. And I think we need to step back and consider KriKri's expectations on data models soon.  While the Mapping DSL and the harvesters are model-agnostic, a lot of the higher-level stuff we've done in the last few weeks has hard dependencies on `DPLA::MAP::Aggregation`, this now depends on `DPLA::MAP::Agent`, and some things make assumptions about predicates used.

app/views/krikri/qa_reports/index.html.erb
```diff
@@ -0,0 +1,11 @@
+<% # krikri/qa_reports/index %>
+
+<h1>QA Reports</h1>
```
anarchivist: from here downward should be wrapped in `<div id='main-container' class='container'>...</div>`

app/views/krikri/qa_reports/show.html.erb
```diff
@@ -0,0 +1 @@
+<% # krikri/qa_records/show %>
```
anarchivist: typo - should be `qa_reports`, not `qa_records`

spec/controllers/validation_reports_controller_spec.rb
```diff
@@ -4,13 +4,12 @@
...
-  describe 'GET #index' do
-    login_user
-
-    it 'redirects to /krikri/report_lists#index in absence of valid params' do
-      get :index
-      expect(response).to redirect_to('/krikri/report_lists')
-    end
-  end
-
+  # describe '#show' do
```
anarchivist: Are these the specs that needed to get commented out because of the issues with Blacklight?

no-reply: No, those are specs that never got written.  The problem specs are in `reports_controller_spec.rb`.

no-reply: These *may* suffer from the same problem, though...

spec/models/validation_report_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'spec_helper'
```
anarchivist: Does the `ValidationReport` model need any tests on the `#find` and `#all` methods, or should this be removed?

spec/controllers/records_controller_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      require 'pry'
+      binding.pry unless assigns[:document]
+
+      expect(response).to render_template('krikri/records/show')
+    end
+
+    it 'gets document by local name' do
+      get :show, id: doc_id
+
+      expect(assigns[:document][:id]).to eq agg.rdf_subject.to_s
```
anarchivist: As I'm looking into the inconsistencies with these specs, I'm not sure what you're testing here.

spec/controllers/records_controller_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      end
+
+      it 'filters records by provider' do
+        get :index, provider: 'abc'
+        expect(assigns[:response][:response][:docs]).to be_empty
+      end
+    end
+  end
+
+  describe 'GET #show' do
```
anarchivist: So, the failures with the `it 'renders show view'` and `it 'gets document by local name'` examples seem to be arising because the GET requests are not successful - they're returning a 404 instead.

anarchivist: (I determined this by adding lines with `expect(response).to be_successful` *before* the existing `expect(...)` line in those two examples.)

lib/krikri/qa_query_client.rb
```diff
@@ -0,0 +1,154 @@
...
+    TYPE = :aggregation
+    VALUE = :value
+
+    module_function
+
+    ##
+    # Writes a {SPARQL::Client::Query} against the Repository for values
+    # matching at the end of a chain of predicates.
+    #
+    # @example
```
anarchivist: :100: Awesome work with the examples.

spec/controllers/records_controller_spec.rb
anarchivist: This suggests to me that we may be running into a potential issue with a race condition somewhere.

anarchivist:
I think this is looking good. I haven't tested out the functionality locally yet because I've been focused on the review. Hopefully I can do that first thing tomorrow.

spec/controllers/records_controller_spec.rb
no-reply: The test is intended to verify that the document is retrievable by "local name". i.e. that "moomin" returns an object with `#rdf_subject` of `base_uri / "moomin"`.

`assigns[:document][:id]` contains the full URI, while `doc_id` is the local name.  This was failing before/when it was written, then passed after reintroducing a method which had been removed as "unused" (so it absolutely tests a regression we've already run into).  Now it fails again...

spec/controllers/records_controller_spec.rb
no-reply: Yes. The problem, as I understand it, is that Blacklight is returning no documents (even though documents appear to be present). This leads to a `404`.

anarchivist: Could this also be arising because of the session variable being set elsewhere?

no-reply: I don't think so. One of the things that the later work on these branches does is remove the reliance of individual requests on session state.  The requests should be purely RESTful, and only the nav bar dependent on the session state. If that's not the case, then that's a fix I'd want to get in immediately.

no-reply: I believe this is the relevant code: 
https://github.com/dpla/KriKri/pull/134/files#diff-6f00d0a280384162066b6d68e263d167R108

https://github.com/dpla/KriKri/pull/134/files#diff-6f00d0a280384162066b6d68e263d167R118

app/views/krikri/shared/_navigation.html.erb
no-reply: Do we want to just move it up to underneath the `remove_session_provider` link?

Would this menu be improved by dropping the link to the current provider altogether?

AudreyAltman: Yes!  Drop the link to the current provider. :+1:

no-reply: Done!

app/views/krikri/qa_reports/index.html.erb
no-reply: Nice. Fixed.

app/views/krikri/qa_reports/show.html.erb
no-reply: :rocket:

spec/models/validation_report_spec.rb
no-reply: I added some light specs. I'd like to do some refactoring here later, and test better.

no-reply:
I think everything is now addressed.

The test suite now passes, with `sleep` between indexing and sending requests for the controller specs. I propose opening a chore to come up with a better strategy of dealing with the index-dependent specs.  We could webmock the solr requests, or very aggressively mock blacklight, or just try to index on `before(:all)` where possible, and include the `sleep` there.

app/models/krikri/validation_report.rb
```diff
@@ -0,0 +1,54 @@
...
+    def all(&block)
+      # set values from block
+      instance_eval &block if block_given?
+
+      query_params = { :rows => 0,
+                       'facet.field' => REQUIRED_FIELDS,
+                       'facet.mincount' => 10000000,
+                       'facet.missing' => true }
+      query_params[:fq] = "provider_id:\"#{@provider_id}\"" if
+        @provider_id.present?
```
AudreyAltman: When I tested this in my browser, I noticed that reports do not seem to be correctly filtering by provider.  When I load `http://localhost:3000/krikri/validation_reports/sourceResource_title`, I see one record :+1:  When I load `http://localhost:3000/krikri/validation_reports/sourceResource_title?provider=123`, I see no records, but I think I should see one, correct?

I think the problem is that `@provider_id` is a local name, and the solr query param:
```query_params[:fq] = "provider_id:\"#{@provider_id}\""``` 
is trying to make and exact match between the indexed `provider_id` and the local name.  

Maybe try something like:
```provider_uri = RDF::URI(Krikri::Provider.base_uri) / @provider_id```
```query_params[:fq] = "provider_id:\"#{provider_uri}\""```

AudreyAltman: Hold the phones! I just realized that when I load `http://localhost:3000/krikri/validation_reports/sourceResource_title?provider=123` into my browser, I should not see any records, because the record in the test that appears on the `sourceResource_title` validation report has a `provider_id` of `_:b3`.  So things in the test data ARE working as expected.

However, _if_ there were test data in which a record with a valid `provider_id` had, say, a missing value for `sourceResource_title`, you may still run into trouble with that local name, as described above.  It would be worth testing to find out.

no-reply: Thanks. I'll verify that this is working (and tested)!

AudreyAltman: The source of my confusion is this:  When I load `http://localhost:3000/krikri/reports?provider=123` in my browser window, under "Validation Reports", I see "sourceResource_title missing (1 record)".  That is not correct.  It should say "sourceResource_title missing (0 records)".

app/controllers/krikri/providers_controller.rb
anarchivist: Added as https://www.pivotaltracker.com/story/show/91795818

no-reply:
homerun.

PR #135, opened 2015-04-03
Support multivalued node fields in solr index

no-reply:
This allows the solr schema to support documents with multiple rdf nodes for a given property (i.e. multiple creators, places, etc...).

The test checks `provider` particularly. We could improve it across the board by adding more data to DPLA::MAP::Aggregation's factory.


anarchivist:
:sun_with_face:

PR #136, opened 2015-04-04
Allow Krikri::Indexer#run to receive an (unused) activity_uri

anarchivist:
Despite `Krikri::Indexer` using `Krikri::EntityConsumer` to identify the activity for which to index records, `Krikri::Indexer#run` still needs an `activity_uri` parameter to meet the expectations of the interface
for `Krikri::Job#self.run`.

### DIFF
--- a/lib/krikri/job.rb
+++ b/lib/krikri/job.rb
@@ -28,6 +28,7 @@ module Krikri
     # @param activity_uri  the URI of the activity responsible for
     #   generating the resources. Set this to (e.g.) prov:wasGeneratedBy
     def self.run(agent, activity_uri = nil)
+      return agent.run unless activity_uri && agent.method(:run).arity != 0
       agent.run(activity_uri)
     end
   end
--- a/lib/krikri/software_agent.rb
+++ b/lib/krikri/software_agent.rb
@@ -26,14 +26,23 @@ module Krikri
     ##
     # Return an agent name suitable for saving in an Activity.
     # This is the name of the most-derived class upon which this is invoked.
+    #
+    # @return [String]
     # @see Krikri::Activity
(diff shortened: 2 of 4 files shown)


anarchivist:
This addresses the issue I [described in Slack late yesterday](https://dpla.slack.com/archives/tech/p1428102997000546), where indexing jobs were failing after being enquequed. The jobs would fail with an `ArgumentError` when the `#run` method would be invoked (`wrong number of arguments (1 for 0)`).

anarchivist:
Also, as far as I can tell, unless this is merged we won't be able to index records from enqueued jobs at all.

no-reply:
It sounds like we have a problem in `Job`.  Maybe it ought to send empty arguments if `activity_uri` is nil.  I may submit a second commit to this branch doing that--or just adding some shared examples for `SoftwareAgents`, but will merge tomorrow unless I have something I can do in short order.

:+1:

anarchivist:
Yeah, after looking at this further, what I was overlooking was the relevance of `activity_uri` parameter is different depending on the implementation of `SoftwareAgent`. `Krikri::Harvester`, `Krikri::Mapper`, and `Krikri::Enricher` use `activity_uri` to add PROV-O assertions, unless it is `nil`. Admittedly, it's not necessary for `Krikri::Indexer` since it doesn't modify resources in Marmotta.

no-reply:
Does that look better?

anarchivist:
Yep! I started working on something similar, but your solution is a bit more thorough than mine. Do you want to rebase to get rid of my original commit and your reversion out?

no-reply:
Will do.

no-reply:
:squash:`d

PR #137, opened 2015-04-04
Improve error handling for Krikri::Mapper:Agent#run

anarchivist:
Occasionally, a mapping job will fail if it tries to save a record that is `nil`. The error logger handling itself fails, since it tries to call the `rdf_subject`method on an instance of `NilClass`.

This addresses an issue when I was testing my changes for #136. Not quite as high priority, and probably needs better tests and more discussion.

### DIFF
--- a/lib/krikri/mapper.rb
+++ b/lib/krikri/mapper.rb
@@ -133,8 +133,13 @@ module Krikri
             rec.mint_id! if rec.node?
             activity_uri ? rec.save_with_provenance(activity_uri) : rec.save
           rescue => e
-            Rails.logger.error("Error saving record: #{rec.rdf_subject}\n" \
-                               "#{e.message}\n#{e.backtrace}")
+            if rec
+              Rails.logger.error("Error saving record: #{rec.rdf_subject}\n"\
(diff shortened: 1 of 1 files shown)


no-reply:
Do we know what causes the nil records? That seems like a bug in the mapper, right?

anarchivist:
Not yet, which is why I've isolated this into its own PR.

lib/krikri/mapper.rb
```diff
@@ -133,8 +133,13 @@ def run(activity_uri = nil)
             rec.mint_id! if rec.node?
             activity_uri ? rec.save_with_provenance(activity_uri) : rec.save
           rescue => e
-            Rails.logger.error("Error saving record: #{rec.rdf_subject}\n" \
-                               "#{e.message}\n#{e.backtrace}")
+            if rec
+              Rails.logger.error("Error saving record: #{rec.rdf_subject}\n"\
```
markbreedlove: Because this is a `SoftwareAgent`, you should be able to change this to `log :error, "message..."` and get the benefit of tagged logging with a timestamp.

PR #138, opened 2015-04-05
Krikri::Provider#reload query should use a DISTINCT clause

anarchivist:
`Krikri::Provider#reload` builds out the triples associated with the provider through a SPARQL query. When there are more than a small number (say, 100) of mapped records in your repository, it takes
a very long time for it to build out the provider given that `#reload` calls `set_value` multiple times: one for each of the sum of the number of solutions for each mapped record associated with that provider.

### DIFF
--- a/app/helpers/krikri/application_helper.rb
+++ b/app/helpers/krikri/application_helper.rb
@@ -3,7 +3,11 @@ module Krikri
     ##
     # @return [Array<Blacklight::SolrResponse::Facets::FacetItem>]
     def available_providers
-      Krikri::Provider.all(&:reload)
+      Rails.cache.fetch('krikri/application_helper/available_providers',
+                        expires_in: 1.hour,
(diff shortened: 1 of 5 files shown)


anarchivist:
For what it's worth, I'm seeing the impact of this performance issue on staging, which has about 87K records for ESDN in it.

anarchivist:
See https://www.pivotaltracker.com/story/show/91817988

no-reply:
Does this fix the performance issue? It seems like either way, some use of `Rails.cache` would be wise, here.

no-reply:
I think 22433e4 will fix render slowdowns, even as the (actual) number of `Providers` gets very large.

As the commit comment notes, it might be good to clear and reload this cache key asynchronously as part of the ingest process, and bump the timeout *way* up, to avoid ever building this list in response to requests.

anarchivist:
30d0172 should address the last of the performance issues I saw tonight.

PR #139, opened 2015-04-06
Fixup validation report views

anarchivist:
1. Iterate through each field returned in validation reports, now that title and isShownAt are multivalued in Solr
1. Link to compare view using `local_name(doc.id)`
1. Pass `@current_provider` as to the validation report view

### DIFF
--- a/app/controllers/krikri/validation_reports_controller.rb
+++ b/app/controllers/krikri/validation_reports_controller.rb
@@ -18,11 +18,11 @@ module Krikri
     layout 'krikri/application'
 
     def show
-      provider_id = params[:provider]
+      @current_provider = params[:provider]
       page = params[:page]
       per_page = params[:per_page]
       @response = ValidationReport.new.find(params[:id]) do
-        self.provider_id = provider_id
+        self.provider_id = @current_provider
         self.start = page
         self.rows = per_page
       end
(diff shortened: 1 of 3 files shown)


no-reply:
Can you say something about what was broken? I'd like to get a controller and/or view test covering the issue, and am happy to write it if you can summarize.

anarchivist:
For 1), I was seeing something like the following. Note that `doc['isShownAt_id']` was actually `"http://cdm16373.contentdm.oclc.org/cdm/ref/collection/p15281coll12/id/6609"`. 

```
I, [2015-04-05T19:35:39.759092 #14617]  INFO -- : Completed 500 Internal Server Error in 1690ms (ActiveRecord: 7.0ms)
F, [2015-04-05T19:35:39.762222 #14617] FATAL -- :
ActionView::Template::Error (undefined method `http://cdm16373.contentdm.oclc.org/cdm/ref/collection/p15281coll12/id/6609_path' for #<#<Class:0x007f3f5e8e0b08>:0x007f3f5e8e1080>)`
``` 

For 2), the links to the records were actually being generated with the full URIs and not the local name. 

For 3), `@current_provider` wasn't being set, which led it to always being `nil` in the context of the view. (I was cherry picking from another branch, and apparently my spec didn't make it through; it should be there now.)

no-reply:
I've added a chore to improve the tests for these modules here: https://www.pivotaltracker.com/story/show/91860282. 

I don't think your PR makes the test gaps worse, and it's clearly an immediate improvement to functionality, so... :rocket:

PR #140, opened 2015-04-06
Have CouchdbHarvester make batch requests

markbreedlove:
Have `CouchdbHarvester` issue requests to CouchDB in paginated fashion, instead of using a stream.  Removes reliance by `#records` upon `Analysand::StreamingViewResponse`, which was having problems with certain UTF-8 characters.

Fix `CouchdbHarvester#record`, `#record_ids`, and `#count` methods to  exclude CouchDB design documents.

### DIFF
--- a/lib/krikri/harvesters/couchdb_harvester.rb
+++ b/lib/krikri/harvesters/couchdb_harvester.rb
@@ -47,7 +47,8 @@ module Krikri::Harvesters
     ##
     # Return the total number of documents reported by a CouchDB view.
     #
-    # @param opts [Hash]  Analysand `#view' options
+    # @param opts [Hash]  Analysand::Database#view options
+    #                     - view:  database view name
     # @return [Fixnum]
     #
     def count(opts = {})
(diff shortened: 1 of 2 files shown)


markbreedlove:
Ack, hold off on reviewing this!  There is an oversight in the way design documents are handled.

markbreedlove:
Ok, should be good to review again.  I might improve some of the tests, but go ahead.

lib/krikri/harvesters/couchdb_harvester.rb
```diff
@@ -35,34 +36,79 @@ def initialize(opts = {})
...
     # @see Analysand::Viewing
-    # @see Analysand::StreamingViewResponse
+    # @see http://docs.couchdb.org/en/latest/couchapp/views/collation.html#all-docs
     def records(opts = {})
       view = opts[:view] || @opts[:view]
-      client.view(view, include_docs: true, stream: true).docs.lazy.map do |r|
-        @record_class.build(mint_id(r['_id']), r.to_json, 'application/json')
+      limit = opts[:limit] || @opts[:limit]
+
+      en = Enumerator.new do |e|
```
no-reply: I'm thinking about ways we might be able to break this out into constituent methods.  Maybe targeting something like this as a starting place?

```ruby
def records (opts = {})
  view = opts[:view] || @opts[:view]
  limit = opts[:limit] || @opts[:limit]

  paged_rows.map do |row|
    @record_class.build(mint_id(row['doc']['_id']), row['doc'].to_json, 'application/json')
  end
end

def paged_rows(opts)
  en = Enumerator.new do |e|
    rows_retrieved = 0
    total_rows = nil
    loop do
      v = client.view(view, view_opts)
     total_rows ||= v.total_rows
     v.rows.each do |row|
        next if row['id'].start_with('_design')
        e.yield row
     end
    end
  end
end
```

lib/krikri/harvesters/couchdb_harvester.rb
```diff
@@ -20,6 +20,7 @@ def initialize(opts = {})
       super
       @opts = opts.fetch(:couchdb, view: '_all_docs')
       @opts[:view] ||= '_all_docs'
+      @opts[:limit] ||= 10
```
markbreedlove: Is this option clear as `limit`?  Would `batch_size` be better?

Limit makes sense down below in the guts of the CouchDB view request, but someone looking at the external interface to the class might comprehend `batch_size` more readily.

anarchivist: IMO, this should match whatever CouchDB's external interface provides, so I think `:limit` is correct.

lib/krikri/harvesters/couchdb_harvester.rb
```diff
@@ -35,34 +36,89 @@ def initialize(opts = {})
...
+    #                     - view:  database view name
+    # @return [Fixnum]
+    #
     def count(opts = {})
       view = opts[:view] || @opts[:view]
-      client.view(view,
-                  limit: 0,
-                  include_docs: false,
-                  stream: true).total_rows
+      # The count that we want is the total documents in the database minus
```
anarchivist: I like this implementation - very cool, and very clever (in a positive way)!

anarchivist:
OK - I'm :+1: - I'll squash and merge is @markbreedlove is away.

PR #141, opened 2015-04-06
Rewrite validation reports model/controller; add tests

no-reply:
This slightly reworks the `ValidationReports` model, and updates its consumers--adding tests for the functionality along the way.

There are a few places where integration testing/extensive mocking are called out.  These suggest a need for a more extensive rewrite of `ValidationReports` and/or a `ValidationReportBuilder` pattern to free those modules from that internal knowledge.  In the meanwhile, I think this makes everything work. :)

### DIFF
--- a/app/controllers/krikri/validation_reports_controller.rb
+++ b/app/controllers/krikri/validation_reports_controller.rb
@@ -17,16 +17,29 @@ module Krikri
     # ApplicationController.  It uses krikri's application layout:
     layout 'krikri/application'
 
+    ##
+    # Render the show view
+    #
+    # This differs from {Blacklight::CatalogController}'s normal handling of the
+    # index and show views, Giving a paginated list of response documents
+    # instead of showing a single document.
+    #
+    # @todo: consider bringing this in line with Blacklight's approach
     def show
       @current_provider = params[:provider]
-      page = params[:page]
-      per_page = params[:per_page]
-      @response = ValidationReport.new.find(params[:id]) do
-        self.provider_id = @current_provider
-        self.start = page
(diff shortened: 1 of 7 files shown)


app/controllers/krikri/validation_reports_controller.rb
```diff
@@ -19,14 +19,19 @@ class ValidationReportsController < CatalogController
 
     def show
```
anarchivist: Can we add a comment here about how this diverges from how `Blacklight::CatalogController` expects `#index` and `#show` to be used?

spec/models/validation_report_spec.rb
```diff
@@ -26,24 +34,53 @@
...
-      expect(result).to eq :result
+      it 'allows pagination' do
+        subject.rows = '1'
+        page_one_docs = subject.find('sourceResource_title').docs
+        subject.page = '2'
+        expect(subject.find('sourceResource_title').docs)
+          .not_to eq page_one_docs
+      end
+
+      it 'multilies default pagination' do
```
anarchivist: (minor) typo: maybe something like `it 'paginates based on a default number of rows'`

no-reply: fixed.

app/controllers/krikri/validation_reports_controller.rb
no-reply: Done.

anarchivist:
I'm :+1:, if you want to :squash: or have someone else look at the review.

anarchivist:
:volcano:

PR #142, opened 2015-04-08
Update dataProvider field in validation report

no-reply:
This fixes an issue where all items are identified as not having a `dataProvider`.


anarchivist:
:rocket:

PR #144, opened 2015-04-08
Split geographic coordinate strings

markbreedlove:
Split strings given in edm:Place (`DPLA::MAP::Place`) into latitude and longitude.

### DIFF
--- a/lib/krikri/enrichments/split_coordinates.rb
+++ b/lib/krikri/enrichments/split_coordinates.rb
@@ -19,20 +19,21 @@ module Krikri::Enrichments
     # Enrich a `DPLA::MAP::Place' object by splitting the string given
     # in its `lat' or `long'.
     #
-    # @param place [DPLA::MAP::Place]
-    # @return [DPLA::MAP::Place]
+    # place.lat and place.long are ActiveTriples::Terms, we only care
+    # about the first value. If multiple values are given, this enrichment
+    # will remove them.
     #
+    # @param place [DPLA::MAP::Place]
+    #
+    # @return [DPLA::MAP::Place]
     def enrich_value(place)
       return place if !place.is_a? DPLA::MAP::Place
+      return place unless splittable?(place.lat) || splittable?(place.long)
 
-      # place.lat and place.long are arrays of one value if they are assigned.
-
-      return place if place.lat.first && place.long.first  # no split required
-
-      if place.lat.first
+      if place.lat.any?
         latlong = coord_values(place.lat.first)
         assign_latlong!(place, latlong.first, latlong.last)
-      elsif place.long.first
+      elsif place.long.any?
         latlong = coord_values(place.long.first)
         assign_latlong!(place, latlong.last, latlong.first)
       end
@@ -41,15 +42,8 @@ module Krikri::Enrichments
     end
 
     def assign_latlong!(place, lat, long)
-      # We have to assign these one at a time because we can't do this:
-      # place.lat, place.long = [[nil], [nil]]
-      # ... which results in:
-      #     "value must be an RDF URI, Node, Literal, or a valid datatype.
-      #     See RDF::Literal. You provided nil"
-      # ... although this works fine:
-      # place.lat, place.long = [['10.0'], ['11.1']]
-      place.lat = lat
-      place.long = long
+      place.lat = lat if lat
+      place.long = long if long
     end
 
     ##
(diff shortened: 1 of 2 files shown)


anarchivist:
I'm :+1:, other than the :squash:. I'd like @AudreyAltman and/or @no-reply take a gander.

spec/lib/krikri/enrichments/split_coordinates_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      let(:lat) { '40.0' }
+      include_examples 'nonsensical coordinates'
+    end
+
+    context 'with a latitude that splits into too many values' do
+      let(:lat) { '1.0,2.0,3.0' }
+      include_examples 'nonsensical coordinates'
+    end
+
+    context 'with a latitude containing non-decimal characters' do
```
no-reply: I suspect we'll see some data that doesn't include decimal characters, like `51N, 108W`, but I'm happy to punt this until we hit a provider where this applies.

lib/krikri/enrichments/split_coordinates.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    def assign_latlong!(place, lat, long)
+      # We have to assign these one at a time because we can't do this:
+      # place.lat, place.long = [[nil], [nil]]
+      # ... which results in:
+      #     "value must be an RDF URI, Node, Literal, or a valid datatype.
+      #     See RDF::Literal. You provided nil"
+      # ... although this works fine:
+      # place.lat, place.long = [['10.0'], ['11.1']]
+      place.lat = lat
```
no-reply: I haven't verified, but I think `place.lat, place.long = ['10.0', '11.1']` and `place.lat, place.long = [nil, nil]` will both work fine.  This already seems to be assuming that you're only working with the value of `ActiveTriples::Term#first`.

spec/lib/krikri/enrichments/split_coordinates_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'spec_helper'
+
+describe Krikri::Enrichments::SplitCoordinates do
+  it_behaves_like 'a field enrichment'
+
+  context 'with place object' do
+    let(:lat) { nil }
+    let(:long) { nil }
+    let(:place) do
```
no-reply: Minor style... maybe collapse this onto one line?

`let(:place) { build(:place, lat: lat, long: long }`

lib/krikri/enrichments/split_coordinates.rb
```diff
@@ -0,0 +1,73 @@
...
+    #
+    # @param place [DPLA::MAP::Place]
+    # @return [DPLA::MAP::Place]
+    #
+    def enrich_value(place)
+      return place if !place.is_a? DPLA::MAP::Place
+
+      # place.lat and place.long are arrays of one value if they are assigned.
+
+      return place if place.lat.first && place.long.first  # no split required
```
no-reply: This requires that the mapping be to one of `lat` or `long`, but not both.  That's not what we're doing to date.  Would it be better to check whether either value contains a comma?

AudreyAltman:
I have nothing to add to what Tom wrote.  Looking good, @mb.

no-reply:
This looks great & is ready to go once we clarify the mapping pattern referred to here: https://github.com/dpla/KriKri/pull/144#discussion_r28072831

My notion is that looking for the delimiter is better since it makes fewer assumptions about the mapping/input data.

spec/lib/krikri/enrichments/split_coordinates_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      end
+      it 'assigns longitude' do
+        expect(subject.enrich_value(place).long).to eq expected_long
+      end      
+    end
+
+    shared_examples 'nonsensical coordinates' do
+      it 'returns empty coordinates' do
+        rv = subject.enrich_value(place)
+        expect(rv.lat).to eq []
```
no-reply: I'm now thinking that this should retain the original values. I don't know if we want to delete coordinates that we can't recognize, but if/when we do, that seems like another enrichment. 

Thoughts @anarchivist?

no-reply:
My changes allow mapping to both fields, giving preference to the `lat, long` format if both are given.  They also avoid destroying data unless we have successfully split the string.

no-reply:
@AudreyAltman and @anarchivist just a quick ping to remind that there are a few commits added to this PR yesterday. If one of you can review the latest changes, i think we're ready to :ship:

anarchivist:
:ok_hand:

no-reply:
Squashed.

Squash

anarchivist:
[](http://media.giphy.com/media/74eZarMmBLk8E/giphy.gif)

PR #146, opened 2015-04-10
Add DCMI Type Enrichment

no-reply:
This enrichment uses White Similarity measures to attempt to match hashes to DCMI Type values, building the result.


no-reply:
Actually, hold off on this. You're welcome to review, but I'm going to change the behavior of the no-match case and add that + some of the other type handling stuff as different enrichments.

PR #148, opened 2015-04-10
Adds an enrichment to copy non-DCMI type values

no-reply:
Copies non-DCMI Type values to a new field via a 'generic' enrichment. This allows, e.g. moving values from `dctype` to `dcformat` before clearing the invalid values from `dctype`.


no-reply:
This should be in for ESDN enrichment.

anarchivist:
:+1:

PR #149, opened 2015-04-10
Add `ConvertToSentenceCase` enrichment

no-reply:
Converts a string to sentence case, removing capital letters in the sentences.


no-reply:
This is targeted for the UW enrichment.

PR #150, opened 2015-04-10
Add bnode Deduplication enrichment

no-reply:


no-reply:
This should be in for ESDN enrichment.

lib/krikri/enrichments/dedup_nodes.rb
```diff
@@ -0,0 +1,45 @@
...
+  ##
+  # Enrichment to remove duplicate blank nodes from a Graph, where "duplicate"
+  # means having the same `providedLabel`.
+  class DedupNodes
+    include Krikri::FieldEnrichment
+
+    ##
+    # @param value [Object]
+    #
+    # @return [Object] the original value altered to remove nodes with the same
```
anarchivist: add an `@example`?

anarchivist:
I'm :+1: if we can improve documentation with an `@example`.

no-reply:
:+1: will do.

no-reply:
Doc added and squashed.

PR #151, opened 2015-04-11
Adds a `#last_value` to `ParserValueArray`

no-reply:
This behaves similarly to `#first_value`, in that it avoids overriding `#first` on the underlying (delegated) `Array`, but gives access to a `ParserValueArray` with only the requested values`.


no-reply:
This is needed for the UW mapping.

anarchivist:
:tada:

PR #152, opened 2015-04-13
Add Enrichment instance caching to Enricher

no-reply:
Creates a cache of enrichment objects for the current instance of `Krikri::Enricher`. This allows individual enrichment instances to maintain their own caches/persistence/optimizations for the life of (e.g.) a queued job.

### DIFF
--- a/lib/krikri/enricher.rb
+++ b/lib/krikri/enricher.rb
@@ -83,7 +83,7 @@ module Krikri
     #
     def chain_enrichments!(agg)
       chain.keys.each do |e|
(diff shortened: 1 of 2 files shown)


anarchivist:
I'm :+1: - anyone else want to take a look?

markbreedlove:
Sure, that makes sense.  The frequent instantiation of that class must have been expensive.  I'm happy to merge it.

no-reply:
:rocket: It's not crucial, but it wouldn't hurt to have this in when the UW enrichment starts.

PR #153, opened 2015-04-14
Extend Timeout and Keep Alive for SPARQL::Client

no-reply:
Some of our more intensive SPARQL Queries for reports take on the order of two minutes for Marmotta to query and build the response. This keeps `SPARQL::Client` from timing out on these queries.

### DIFF
--- a/lib/krikri/engine.rb
+++ b/lib/krikri/engine.rb
@@ -75,6 +75,11 @@ module Krikri
     initializer :rdf_repository do
       Krikri::Repository =
         RDF::Marmotta.new(Krikri::Settings['marmotta']['base'])
+
+      Krikri::Repository.query_client.instance_eval do
+        @http.keep_alive = 360
+        @http.read_timeout = 360
+      end
     end
 
     initializer :blacklight_settings do


no-reply:
Hold on this. Apparently 360 seconds isn't long enough. :/

lib/krikri/engine.rb
```diff
@@ -75,6 +75,11 @@ def configure_blacklight!
     initializer :rdf_repository do
       Krikri::Repository =
         RDF::Marmotta.new(Krikri::Settings['marmotta']['base'])
+
+      Krikri::Repository.query_client.instance_eval do
+        @http.keep_alive = 360
```
jeremyf: Have you considered using Figaro? Or dotenv? This would be to set an application configuration variable.

https://www.ruby-toolbox.com/categories/Configuration_Management

lib/krikri/engine.rb
```diff
@@ -75,6 +75,11 @@ def configure_blacklight!
     initializer :rdf_repository do
       Krikri::Repository =
         RDF::Marmotta.new(Krikri::Settings['marmotta']['base'])
```
jeremyf: Would this be better as `.fetch` so you have proper key errors if the settings are wonky?

lib/krikri/engine.rb
no-reply: We're using `rails_config`, which is working well enough for us.  You're right that this should be a configuration variable, though.  (Would also be nice if `SPARQL::Client` offered a better way of tweaking the `keep_alive` than `instance_eval`).

lib/krikri/engine.rb
no-reply: This rung true to me, but i guess not: 
```Krikri::Settings.fetch('blah')
NoMethodError: undefined method `fetch' for #<RailsConfig::Options ...```

jeremyf: Alas :disappointed:

PR #154, opened 2015-04-16
Format Aggregations to MAP 3.1 for indexing

no-reply:
Adds a `CrosswalkHashBuilder` that constructs DPLA MAP 3.1 JSON-LD, with appropriate additions for indexing into our production ElasticSearch.

### DIFF
--- a/lib/krikri/map_crosswalk.rb
+++ b/lib/krikri/map_crosswalk.rb
@@ -1,0 +1,252 @@
+module Krikri
+  ##
+  # Adds ability to generate a DPLA MAPv3.1 compliant JSON-LD
+  # representation to `DPLA::MAP::Aggregation`.
+  module MapCrosswalk
+    ##
+    # @return [Hash] a DPLA MAP 3.1 (APIv2) object
+    def to_3_1_json
+      CrosswalkHashBuilder.build(self)
+    end
+
+    ##
+    # Generates a MAPv3.1 (APIv2) hash for the given parent Aggregation
+    #
+    # @example
+    #   builder = CrosswalkHashBuilder.new(agg)
+    #   builder.build # => {...}
+    #
+    # @example
+    #   CrosswalkHashBuilder.build(agg) # => {...}
+    #
+    class CrosswalkHashBuilder
+      attr_accessor :hash
+
+      ##
+      # @param parent [DPLA::MAP::Aggregation]
+      def initialize(parent)
+        @parent = parent
+        @hash = {}
+      end
+
+      class << self
+        def build(parent)
+          new(parent).build
+        end
+      end
+
+      def build
+        @hash = {}
+        build_aggregation
+      end
+
+      private
+
+      ##
+      # @todo: FIXME! The legacy system increments `ingestionSequence` for each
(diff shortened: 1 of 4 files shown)


anarchivist:
I'm going to leave comments here, which generally relate to the Elasticsearch indexing and may (but not necessarily) relate to the PR itself.

anarchivist:
With UW records, for some reason it appears the `skos:prefLabel` property isn't getting populated. @no-reply's [gist from last night](https://gist.github.com/no-reply/0daeef87d53a432d0475) shows this same behavior. It's weird, because they are getting populated for the ESDN mapping: https://gist.github.com/anarchivist/d132be618b45e2b4c231

anarchivist:
This gist contains examples of dumped Turtle UW and ESDN records: https://gist.github.com/anarchivist/d132be618b45e2b4c231

lib/krikri/map_crosswalk.rb
```diff
@@ -0,0 +1,234 @@
...
+          lang.prefLabel.any? ? lang.prefLabel : lang.providedLabel
+        end
+
+        set_value(sr, :publisher, parent_sr.publisher) do |pub|
+          pub.label.any? ? pub.label : pub.providedLabel
+        end
+        set_value(sr, :relation, parent_sr.relation)
+
+        set_value(sr, :genre, parent_sr.genre) do |genre|
+          genre.prefLabel.any? ? genre.prefLabel : genre.providedLabel
```
anarchivist: For `genre` and others, the expectation that it's a graph is making the mapping somewhat brittle, leading to errors like `Indexing error: undefined method `prefLabel' for "Photograph; b&w":String`. I'm guessing we should probably have some logic to handle strings, even though we know that the longer term goal is to have these properties behave like we expect.

no-reply: I'm inclined to skip values that don't respond to the expected methods.  That seems safer than blindly mapping whatever string we get, yes?

anarchivist: I think that's OK, as long as we log that somewhere.

anarchivist:
I think this is largely looking good, but as my comments suggest, the biggest issue is the brittleness when incoming graphs don't follow the expectations of the model closely. As long as we can handle the exceptions sanely and log any transformation errors so we can diagnose indexing issues later. The major issue in this case is that the indexer jobs die, but are seen as successful, if any exceptions get raised.

no-reply:
The lastest updates make this much less fragile (see `#get_label`).  It should now handle cases where a string appears where we're expecting a resource.  Depending on the representation in 3.1, we either map those strings directly or ignore them (e.g. for `:provider`).

I think this is now ready to go, though, as you say, we may need to fix the error handling in `Indexer` ASAP.

anarchivist:
There are three issues that we need to address from where I sit before this is mergeable:

1) `place` should be `spatial`
2) `provider` should be an object and not an array, because of an implementation issue in the frontend: https://github.com/dpla/frontend/blob/develop/app/models/item.rb#L14
3) After enrichment(?) `subject[].name` should be a string and not an array (minor frontend aesthetic issue): 

screen shot 2015-04-16 at 1 14 15 pm

PR #155, opened 2015-04-16
Add enrichment to split a resource by its providedLabel

no-reply:
This is needed to split the UW subjects.

### DIFF
--- a/lib/krikri/enrichments/split_provided_label_at_delimiter.rb
+++ b/lib/krikri/enrichments/split_provided_label_at_delimiter.rb
@@ -1,0 +1,78 @@
+module Krikri::Enrichments
+  ##
+  # Splits a Resource into multiple resources of its own class by a given
+  # delimiter in its `#providedLabel`. The full original resource is retained
+  # with the first value from the split label.
+  #
+  # @example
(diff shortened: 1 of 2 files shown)


lib/krikri/enrichments/split_provided_label_at_delimiter.rb
```diff
@@ -0,0 +1,78 @@
...
+      resource.providedLabel = providedLabel
+      resource
+    end
+
+    ##
+    # @param value [#providedLabel]
+    # @return [Array<String>] a flat array of provided labels split from
+    #   the original
+    def split_value(value)
+      value.providedLabel.map { |l| l.split(';').map(&:strip) }.flatten
```
AudreyAltman: Should this split by `@delimiter` instead of `;`?

no-reply: yes!

AudreyAltman:
Aside from the one comment above, this looks good to me.  It may be good for @anarchivist or @markbreedlove to take a look too, since they are more familiar with enrichments.

anarchivist:
:+1:

PR #156, opened 2015-04-16
Update Crosswalk mapping for preview/object change

no-reply:
`edm:object` was used in MAPv3.x to mean `edm:preview`. That's fixed in 4.0; this fixes it in the mapping.


anarchivist:
:moneybag:

PR #157, opened 2015-04-21
Add additional facets and change pagination

anarchivist:
This is still in progress, but I'm opening a PR since I'll need some help with someone (maybe @AudreyAltman) on this - I'm having some JS issues that are preventing the modal from opening.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -4,7 +4,18 @@ Krikri::Engine.routes.draw do
   # TODO: remove unnecessary :harvest_sources and :institutions routes once we
   # have established what we do and don't need
   resources :validation_reports, only: [:show]
-  resources :records, only: [:index, :show]
+
+  # Blacklight::CatalogController subclasses seem to have an issue where their
+  # facet-related routes don't get automatically populated, leading to routing
+  # errors when facet limits are enabled and the view attempts to render the
+  # "more" links. This fix was suggested by Trey Terrell, demonstrated here:
+  # https://github.com/OregonDigital/oregondigital/blob/d1653e41/config/routes.rb#L52-L56
+  resources :records, only: [:index, :show] do
+    collection do
+      get "facet/:id", :to => "records#facet"
+    end
+  end
+
   resources :reports, only: [:index]
   resources :qa_reports, only: [:index, :show]
   resources :institutions do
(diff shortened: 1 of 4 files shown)


anarchivist:
Here's the JS error I'm running into: 
```
Uncaught TypeError: $(...).modal is not a function
Blacklight.ajaxModal.receiveAjax @ ajax_modal.js?body=1:141
fire @ jquery.js?body=1:3120
self.fireWith @ jquery.js?body=1:3232
done @ jquery.js?body=1:9278
send.callback @ jquery.js?body=1:9686
```

anarchivist:
This should be ready, but I'd like to review aesthetics and usability with @guegueng before we merge.

anarchivist:
OK - this is ready.

solr_conf/solrconfig.xml
```diff
@@ -140,7 +140,7 @@
        <str name="facet">true</str>
        <str name="facet.mincount">1</str>
        <str name="facet.limit">10</str>
-       <str name="facet.field">dataProvider_name</str>
+       <str name="facet.field">dataProvider_providedLabel</str>
```
AudreyAltman: The above two lines could be deleted.  I think I forgot to delete them way back when I was learning how solr configuration worked.  They don't hurt anything, either, but I don't think they ever get used.

anarchivist: I think they define the default behavior for the Solr `SearchHandler`, meaning that automatically it would return `dataProvider_providedLabel` as a facet if no other facets are explicitly requested.

config/routes.rb
```diff
@@ -4,7 +4,11 @@
   # TODO: remove unnecessary :harvest_sources and :institutions routes once we
   # have established what we do and don't need
   resources :validation_reports, only: [:show]
-  resources :records, only: [:index, :show]
+  resources :records, only: [:index, :show] do
+    collection do
+      get "facet/:id", :to => "records#facet"
+    end
+  end
```
AudreyAltman: Could you add a comment explaining this change?

AudreyAltman:
I tested this locally and it is working as expected.  Other than my two small suggestions above, I think this is ready to merge.

solr_conf/solrconfig.xml
AudreyAltman: Yes, I believe you are correct.

anarchivist:
Documentation added, and :squash:ed.

AudreyAltman:
:+1: this look ready to :merge:

PR #158, opened 2015-04-27
remove scroll from side-by-side records

AudreyAltman:
This removes the scrolling feature for individual records on the side-by-side record comparison.


markbreedlove:
This looks good, and I can confirm that it works as planned in my local environment.  I'm willing to merge it if nobody else has any further observations.

no-reply:
:rocket:

PR #159, opened 2015-04-27
Test support for Ruby versions

no-reply:
Add previous stable, 2.0.0 stable, and current stable to Travis config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,11 @@
 language: ruby
 bundler_args: --without debug
 script: "bundle exec rake ci"
-cache: bundler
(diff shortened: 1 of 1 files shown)


anarchivist:
After months of tinkering with this, it seems like we were able to get this working. 2.0.0 is failing, so after discussion with @no-reply we decided that we were only going to support 2.1+. 

The previous build errors on 2.1.6 and 2.2.x seem to stem from the same issue as travis-ci/travis-ci#4822, so the commit also now disables caching of dependencies. This will slow down the build process a little bit, not impossibly so.

PR #160, opened 2015-04-27
Change index fields for records

AudreyAltman:
This adds 'Creator', 'Subject', and 'Collection' to the index fields for a record on the search results page.  It also reorders the fields.  It changes the source field for 'Type' from `sourceResource.type.id` to `sourceResource.type.name`. 

To test locally, generate a test app and start jetty.  You may need to stop, clean, and config jetty first:
`bundle exec rake jetty:stop`
`bundle exec rake jetty:clean`
`bundle exec rake jetty:config`
`bundle exec rake jetty:start`

Index a sample record with a value for `sourceResource.type.name`.  In your rails console:
`include FactoryGirl::Syntax::Methods`
`require 'dpla/map/factories'`
`agg = build(:aggregation)`
`agg.mint_id!('krikri_sample')`
`indexer = Krikri::QASearchIndex.new`
`graph = agg.to_jsonld['@graph'].first`
`graph['sourceResource']['type']['name'] = "Image"`
`indexer.add graph`
`indexer.commit`

### DIFF
--- a/solr_conf/schema.xml
+++ b/solr_conf/schema.xml
@@ -631,6 +631,7 @@ v         currently supported on types that are sorted internally as strings
    <field name="sourceResource_type_id" type="string" indexed="true" stored="true" multiValued="true" />
    <field name="sourceResource_type_type" type="string" indexed="false" stored="false" multiValued="true" />
    <field name="sourceResource_type_providedLabel" type="string" indexed="true" stored="true" multiValued="true" />
+   <field name="sourceResource_type_name" type="string" indexed="true" stored="true" multiValued="true" />
 
    <!-- Dynamic field definitions.  If a field name is not found, dynamicFields
         will be used if the name matches any of the patterns.
@@ -735,6 +736,7 @@ v         currently supported on types that are sorted internally as strings
   <copyField source="sourceResource_subject_providedLabel" dest="text" />
   <copyField source="sourceResource_temporal" dest="text" />
   <copyField source="sourceResource_title" dest="text" />
+  <copyField source="sourceResource_type_name" dest="text" />
 
    <!-- Above, multiple source fields are copied to the [text] field.
 	  Another way to map multiple source fields to the same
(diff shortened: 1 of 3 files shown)


solr_conf/schema.xml
```diff
@@ -631,6 +631,7 @@ v         currently supported on types that are sorted internally as strings
    <field name="sourceResource_type_id" type="string" indexed="true" stored="true" multiValued="true" />
    <field name="sourceResource_type_type" type="string" indexed="false" stored="false" multiValued="true" />
    <field name="sourceResource_type_providedLabel" type="string" indexed="true" stored="true" multiValued="true" />
+   <field name="sourceResource_type_name" type="string" indexed="true" stored="true" multiValued="true" />
```
anarchivist: Where are these `*_name` fields getting their data?

anarchivist:
I think this probably has some connection with how we want to address the indexing behavior in [#93228766](https://www.pivotaltracker.com/story/show/93228766).

solr_conf/schema.xml
AudreyAltman: You may need to talk to @guegueng about this.  I deferred to her knowledge of MAP when articulating the field names.

no-reply: The JSON-LD context maps `name` to `skos:prefLabel`. https://github.com/dpla/dpla_map/blob/map-4.0/lib/dpla/map/context.rb#L63

AudreyAltman: Is the field `sourceResource.type.name` currently being indexed into Solr?  Because if not, I could use `sourceResource.type.providedLabel` or `sourceResource.type.id` for this PR.

no-reply: I'm not sure what the question is... I may be missing some important context.

I think it will be indexed if there is a schema entry for it, and not otherwise, right? https://github.com/dpla/KriKri/blob/develop/lib/krikri/search_index.rb#L266-L269

no-reply: Conversation on this follows from: https://dpla.slack.com/archives/tech/p1430232893000121

Specifically, there's some question to resolve about how we use `prefLabel` and `providedLabel`, and how they should be indexed in this case.

no-reply: Further:

anarchivist:
I'm fine with this, all said.

PR #162, opened 2015-05-26
Add ISO enrichment

no-reply:
Allows matching on Lexvo ISO 639-3 URIs. Matches string values and blank nodes with `providedLabel` properties. For the first match against an exact ISO code (or a label, e.g. 'english' if no code matches', produces a new blank node with `providedLabel` set to the input value and a `skos:exactMatch` link to the Lexvo URI.

### DIFF
--- a/lib/krikri/enrichments/language_to_lexvo.rb
+++ b/lib/krikri/enrichments/language_to_lexvo.rb
@@ -1,0 +1,161 @@
+module Krikri::Enrichments
+  ##
+  # Converts text fields and/or providedLabels to ISO 639-3 URIs (Lexvo)
+  #
+  # Transforms text values matching either (English) labels or language codes
+  # from Lexvo into DPLA::MAP::Controlled::Language resources with
+  # skos:exactMatch of the appropriate Lexvo URIs. Original string values are
+  # retained as dpla:providedLabel.
+  #
+  # Currently suppports langague codes in ISO 639-3, but may be extended to
+  # matchother two and three letter codes in Lexvo with ISO 639-3 URIs.
+  #
+  # @example matching string values
+  #   iso = LanguageToLexvo.new
+  #   iso.enrich_value('fin')     # matches 'http://lexvo.org/id/iso639-3/fin'
+  #   iso.enrich_value('finnish') # matches 'http://lexvo.org/id/iso639-3/fin'
+  #   iso.enrich_value('eng')     # matches 'http://lexvo.org/id/iso639-3/eng'
+  #   iso.enrich_value('english') # matches 'http://lexvo.org/id/iso639-3/eng'
+  #   iso.enrich_value('English') # matches 'http://lexvo.org/id/iso639-3/eng'
+  #
+  # If no matches are found, returns a bnode with the input value as
+  # providedLabel.
+  #
+  # If passed an ActiveTriples::Resource, the enrichment will:
+  #
+  #   - Perform the above text matching on any present `providedLabel`s,
+  #     returning the original node if no results are found.  If multiple
+  #     values are provided and multiple matches found, they will be
+  #     deduplicated.
+  #   - Leave DPLA::MAP::Controlled::Language objects that are not bnodes
+  #     unaltered.
+  #   - Remove any values which are not either bnodes or members of
+  #     DPLA::MAP::Controlled::Language.
+  #
+  # @example matching node values
+  #   iso = LanguageToLexvo.new
+  #   lang = DPLA::MAP::Controlled::Language.new
+  #   lang.providedLabel = 'eng'
+  #   iso.enrich_value(lang)     # matches ['http://lexvo.org/id/iso639-3/fin']
+  #   lang.providedLabel = 'fin', 'eng'
+  #   iso.enrich_value(lang)     # matches ['http://lexvo.org/id/iso639-3/fin',
+  #                              #          'http://lexvo.org/id/iso639-3/eng']
+  #
+  # Label matches are cached within the enrichment instance, 
+  #
+  # @see DPLA::MAP::Controlled::Language
+  # @see http://www.lexvo.org/
+  class LanguageToLexvo
+    include Krikri::FieldEnrichment
+
+    TERMS = DPLA::MAP::Controlled::Language.list_terms.freeze
+    QNAMES = TERMS.map { |t| t.qname[1] }.freeze
+
+    ##
+    # Runs the enrichment against a node. Can match literal values, and Language
+    # values with a provided label.
+    #
+    # @example with a matching value
+    #   lang = enrich_value('finnish')
+    #   #=> #<DPLA::MAP::Controlled::Language:0x3f(default)>
+    #   lang.providedLabel
+    #   #=> ['finnish']
+    #   lang.exactMatch.map(&:to_term)
+    #   #=> [#<RDF::Vocabulary::Term:0x9b URI:http://lexvo.org/id/iso639-3/fin>]
+    #
+    # @example with no match
+    #   lang = enrich_value('moomin')
+    #   #=> #<DPLA::MAP::Controlled::Language:0x3f(default)>
+    #   lang.providedLabel
+    #   #=> ['moomin']
+    #   lang.exactMatch
+    #   #=> []
+    #
+    # @param value [ActiveTriples::Resource, #to_s]
+    # @return [DPLA::MAP::Controlled::Language, nil] a resource representing the
+    #   language match.
+    def enrich_value(value)
+      return enrich_node(value) if value.is_a?(ActiveTriples::Resource) &&
+        value.node?
+      return value if value.is_a?(DPLA::MAP::Controlled::Language)
+      return nil if value.is_a?(ActiveTriples::Resource)
+      enrich_literal(value)
+    end
+
+    ##
+    # Runs the enrichment over a specific node, accepting an
+    # `ActiveTriples::Resource` with a provided label and returning a new node
+    # with a lexvo match.
+    #
+    # @param value [ActiveTriples::Resource] a resource with a
+    #   `dpla:providedLabel`
+    # @return [Array<ActiveTriples::Resource>, ActiveTriples::Resource]
+    def enrich_node(value)
+      labels = value.get_values(RDF::DPLA.providedLabel)
+      return value if labels.empty?
+      labels.map { |label| enrich_literal(label) }
+    end
+
+    ##
+    # Runs the enrichment over a string.
+    #
+    # @param label [#to_s] the string to match
+    # @return [ActiveTriples::Resource] a blank node with a `dpla:providedLabel`
+    #   of `label` and a `skos:exactMatch` of the matching lexvo language,
+    #   if any
+    def enrich_literal(label)
+      node = DPLA::MAP::Controlled::Language.new()
+      node.providedLabel = label
+
+      match = match_iso(label.to_s)
+      match = match_label(label.to_s) if match.node?
+
+      node.exactMatch = match unless match.node?
+
+      node
+    end
+
+    ##
+    # Converts string or symbol for a three letter language code to an
+    # `ActiveTriples::Resource`.
+    #
+    # @param code [#to_sym] a three letter iso code
+    # @return [DPLA::MAP::Controlled::Language]
+    def match_iso(code)
+      match = QNAMES.find { |c| c == code.to_sym }
+      from_sym(match)
+    end
+
+    ##
+    # Converts string or symbol for a language label to an
+    # `ActiveTriples::Resource`.
+    # 
+    # Matched values are cached in an instance variable `@lang_cache` to avoid
+    # multiple traversals through the vocabulary term labels.
+    #
+    # @param code [#to_sym] a string to match against a language label
+    # @return [DPLA::MAP::Controlled::Language]
+    def match_label(label)
+      @lang_cache ||= {}
+      return @lang_cache[label] if @lang_cache.keys.include? label
+
+      match = TERMS.find do |t|
+        Array(t.label).map(&:downcase).include? label.downcase
+      end
+      
+      # Caches and returns the the label match
+      @lang_cache[label] = from_sym(match)
+    end
+
+    private
+
+    ##
+    # @param code [#to_s] A language code to convert to a URI
+    # @return [DPLA::MAP::Controlled::Language] a language Resource with the 
+    #   matching code as the URL's local name. this will be a Node if code is
+    #   `nil`
+    def from_sym(code)
+      DPLA::MAP::Controlled::Language.new(code)
+    end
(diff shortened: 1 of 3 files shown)


lib/krikri/enrichments/iso_enrich.rb
```diff
@@ -0,0 +1,150 @@
...
+  #   iso.enrich_value(lang)     # matches ['http://lexvo.org/id/iso639-3/fin']
+  #   lang.providedLabel = 'fin', 'eng'
+  #   iso.enrich_value(lang)     # matches ['http://lexvo.org/id/iso639-3/fin',
+  #                              #          'http://lexvo.org/id/iso639-3/eng']
+  #
+  # Label matches are cached within the enrichment instance.
+  #
+  # @see DPLA::MAP::Controlled::Language
+  # @see http://www.lexvo.org/
+  class IsoEnrich
```
anarchivist: I'd rename the class something slightly clearer, e.g. `LanguageToISO639` - as is, the class name isn't obvious since ISO provides a number of standards we'll use in the application.

lib/krikri/enrichments/iso_enrich.rb
```diff
@@ -0,0 +1,150 @@
+module Krikri::Enrichments
+  ##
+  # Converts text fields and/or providedLabels to ISO 689-3 URIs (Lexvo)
```
anarchivist: typo: `ISO 639-3`

lib/krikri/enrichments/iso_enrich.rb
```diff
@@ -0,0 +1,150 @@
...
+  # @see DPLA::MAP::Controlled::Language
+  # @see http://www.lexvo.org/
+  class IsoEnrich
+    include Krikri::FieldEnrichment
+
+    TERMS = DPLA::MAP::Controlled::Language.list_terms.freeze
+    QNAMES = TERMS.map { |t| t.qname[1] }.freeze
+
+    ##
+    # Runs the enrichement against a node. Can match literal values, and Language
```
anarchivist: Typo: `enrichment`

lib/krikri/enrichments/iso_enrich.rb
```diff
@@ -0,0 +1,150 @@
...
+      match = QNAMES.find { |c| c == code.to_sym }
+      from_sym(match)
+    end
+
+    ##
+    # Converts string or symbol for a language label to an
+    # `ActiveTriples::Resource`.
+    #
+    # @param code [#to_sym] a string to match against a language label
+    # @return [DPLA::MAP::Controlled::Language]y
```
anarchivist: typo: extra `y` at end of line

no-reply: emacs artifacts. :(

anarchivist: [Ed is the standard editor.](http://www.gnu.org/fun/jokes/ed-msg.html)

lib/krikri/enrichments/iso_enrich.rb
no-reply: Good point. Might go for `LanguageToLexvo` or something like, since that's slightly more specific and probably easier to work with in code.

lib/krikri/enrichments/iso_enrich.rb
```diff
@@ -0,0 +1,150 @@
...
+      node
+    end
+
+    ##
+    # Converts string or symbol for a three letter language code to an
+    # `ActiveTriples::Resource`.
+    #
+    # @param code [#to_sym] a three letter iso code
+    # @return [DPLA::MAP::Controlled::Language]
+    def match_iso(code)
```
anarchivist: This is an aside, because I'm pretty sure that it wasn't in scope for the original story, but it might be useful. The old ingestion system's [`enrich_language` module](https://github.com/dpla/ingestion/blob/develop/lib/akamod/enrich_language.py) also mapped ISO 639-1 (alpha-2) codes to ISO 639-3 codes. I don't think I've seen these rampant in any data recently, but it's worth knowing about.

no-reply: Yes, I believe that's deliberately left out of scope, for now. I'd propose to reintroduce it if/when it comes up.  It might be worth softening the language on the class-level documentation slightly to allow that this could/would be the place to convert other language formats to Lexvo.

anarchivist: :+1: Duly noted.

anarchivist:
Looking good. A few related comments:

* Are we concerned about mapping issues w/ ISO 639-2 and the MARC language codes list? For example, the MARC language code for "Chinese" is `chi`, while in ISO 639-3 it would be `zho`.  It's worth noting that Lexvo now seems to have incorporated the MARC code mappings.
* For some reason we broke the [ISO 639-3 mappings](https://github.com/dpla/ingestion/blob/develop/lib/iso639_3.py) for the old system into two chunks, `ISO_639_3_1` and `ISO_639_3_2`. I think this had something to do with regex matching - perhaps https://issues.dp.la/issues/7618.

I'll tag @markbreedlove to give this a look as well.

anarchivist:
> It's worth noting that Lexvo now seems to have incorporated the MARC code mappings.

... which means we might want to regenerate the `RDF::Vocabulary` for the Lexvo ISO 639-3 codes for `DPLA::MAP`.

no-reply:
I think the latest deals with the various comments.

We agreed to punt on two-letter and MARC codes for now. I added a comment to the enrichment to reflect that.

I expect regex and other fuzzy matching will either be a separate enrichment, or will grow out of this as required to deal with real data.

anarchivist:
:cn: :fr: :100:

anarchivist:
I'm good on this, but as noted, I'd like someone else to take a look. Otherwise, :squash: :jack_o_lantern:.

no-reply:
Rebased.  I'll defer on :merge: for further review, if anyone can jump in.

lib/krikri/enrichments/language_to_lexvo.rb
```diff
@@ -0,0 +1,153 @@
...
+  # matchother two and three letter codes in Lexvo with ISO 639-3 URIs.
+  #
+  # @example matching string values
+  #   iso = LanguageToLexvo.new
+  #   iso.enrich_value('fin')     # matches 'http://lexvo.org/id/iso639-3/fin'
+  #   iso.enrich_value('finnish') # matches 'http://lexvo.org/id/iso639-3/fin'
+  #   iso.enrich_value('eng')     # matches 'http://lexvo.org/id/iso639-3/eng'
+  #   iso.enrich_value('english') # matches 'http://lexvo.org/id/iso639-3/eng'
+  #   iso.enrich_value('English') # matches 'http://lexvo.org/id/iso639-3/eng'
+  #   iso.enrich_value('English') # matches 'http://lexvo.org/id/iso639-3/eng'
```
markbreedlove: It looks like this was repeated.

no-reply: Good catch.

lib/krikri/enrichments/language_to_lexvo.rb
```diff
@@ -0,0 +1,153 @@
...
+      @lang_cache[label] = from_sym(match)
+    end
+
+    private
+
+    ##
+    # @param code [#to_s] A language code to convert to a URI
+    # @return [DPLA::MAP::Controlled::Language]
+    def from_sym(code)
+      return nil if code.nil?
```
markbreedlove: Do you think this is OK to return nil if the documented return type is `DPLA::MAP::Controlled::Language`?

no-reply: No. You're right.

My impulse for a method like this is to fix the documentation:
`# @return [DPLA::MAP::Controlled::Language, nil] the language Resource generated from the symbol; nil if a nil code is given.`

There's probably a way to make the possibility of a `nil` return value unnecessary, and that would be better. Ideas?

lib/krikri/enrichments/language_to_lexvo.rb
```diff
@@ -0,0 +1,153 @@
...
+    # @return [DPLA::MAP::Controlled::Language]
+    def match_label(label)
+      @lang_cache ||= {}
+      return @lang_cache[label] if @lang_cache.keys.include? label
+
+      match = TERMS.find do |t|
+        Array(t.label).map(&:downcase).include? label.downcase
+      end
+
+      @lang_cache[label] = from_sym(match)
```
markbreedlove: This appears to be using the side effect of an assignment.  Is that intentional?

I know it's done a lot in other languages and contexts, but it's not always possible to tell if it is an oversight.

markbreedlove: Side effect may be the wrong term in this case, but I mean, of course, the return value of the assignment itself.

no-reply: Understood.

Yeah, this relies on `Hash#[]=` to return the set value (which is the [documented return value](http://ruby-doc.org/core-2.1.5/Hash.html#method-i-5B-5D-3D)).  I think I considered this as slightly more clear:

```ruby
@lang_cache[label] = from_sym(match)
@lang_cache[label]
```

But I'm not sure that's *actually* more clear, and this is slightly cheaper.

lib/krikri/enrichments/language_to_lexvo.rb
markbreedlove: What if `DPLA::MAP::Controlled::Language` overrode `#empty?`  It could return `true` if it's instantiated with a code of `nil`, or if there are other circumstances that fit.

no-reply: `DPLA::MAP::Controlled::Language` is an `RDF::Enumerable`/`RDF::Countable`, so it has commitments re: `#empty?`: http://www.rubydoc.info/github/ruby-rdf/rdf/RDF/Countable#empty%3F-instance_method

Can you say more about this suggestion, though?

Currently:

```ruby
r = ActiveTriples::Resource.new(nil)
# => #<ActiveTriples::Resource:0x3ff931cf3edc(#<ActiveTriples::Resource:0x007ff2639e7db8>)>
r.rdf_subject
# => #<RDF::Node:0x3ff931cd3538(_:g70339514938680)>
r.node?
# => true
```

markbreedlove: I was just thinking that if it can be instantiated without an argument, as in https://github.com/dpla/KriKri/blob/feature/iso-enrich/lib/krikri/enrichments/language_to_lexvo.rb#L109 then it must be empty in some sense, or you would want to be able to evaluate it as empty or false.  If that's hard or impractical to do, then I don't really have that much to add at the moment.  Having it able to evaluate to false or empty would allow you to pass around the same type of object without worrying about missing methods or whatever else would amount to type conflicts.

no-reply: > must be empty in some sense, or you would want to be able to evaluate it as empty or false.

Yeah, it would be "empty" in the sense that it's a `AT::Resource` with `RDF::Node` rather than an `RDF::URI`.  I don't think that much helps us re: passing the object around unless, as at L109, we're aiming to return a `Node`.

no-reply:
@markbreedlove I think there are two things you've called out for possible adjustments:

  - The maybe too clever use of `Hash#[]=`'s return value. https://github.com/dpla/KriKri/pull/162#discussion_r31650628
    - We could switch this to a two line call for clarity; do we want to adopt that as a general development guideline? Or is the terse usage here preferable?
  - On https://github.com/dpla/KriKri/pull/162#discussion_r31649145 we either need to clean up the documentation a bit or figure out a good solution that avoids the need for `nil`. I have mixed feelings here. I'm generally reticent about `nil` return values, but I'm not sure I have a great alternative.
     - We could return the bnode `Resource` we get back from `#new(nil)`, and let the client code decide whether it wants to keep it.
     - We could use a `Maybe`-style monad.
     - ??? (profit?)
     - Should I take the time to try to strip `nil` handling out of this module with one of these approaches, or is this maybe a longer discussion?

I'm looking to you, @anarchivist, and @AudreyAltman for advice about how to proceed.

anarchivist:
>   - The maybe too clever use of `Hash#[]=`'s return value. https://github.com/dpla/KriKri/pull/162#discussion_r31650628
>    - We could switch this to a two line call for clarity; do we want to adopt that as a general development guideline? Or is the terse usage here preferable?

+1 on terseness, but add documentation that explicitly describes the caching behavior.

>   - On https://github.com/dpla/KriKri/pull/162#discussion_r31649145 we either need to clean up the documentation a bit or figure out a good solution that avoids the need for `nil`. I have mixed feelings here. I'm generally reticent about `nil` return values, but I'm not sure I have a great alternative.

I don't have a strong feeling about it, but I think we should revise the documentation to start.. Can you say more about option 1 (returning the bnode `Resource` returned by `#new(nil)`)? I'm leaning towards this option as being the best. What would be the most consistent approach based on expectations from RDF.rb?

markbreedlove:
It's funny because I think
```
a = b()
a
```
... is more clear because it removes ambiguity and doesn't lead me to go looking for the definition of `#b` to double-check that its return type is the current method's return type.  It seems obvious that the author has vetted the type of `b` and intends to pass it on.   So it's not terse, but...

If only we could evaluate various objects in a boolean context.  That would allow us to be a lot more terse, and we can't do that, because (for example) an empty array is `true`, and a lot of other "empty" things.  (Sad trombone sound.)  Too bad that `nil` evaluates to `false` but `[]` doesn't, even though `nil` is an object just like `[]`.

I don't think I actually want to have strict typing of all return values; I think it would be ideal for every applicable function to be documented with `@return [SomeClass, nil]` or for that object to respond to `empty?`.   Unfortunately, `nil` does not respond to `empty?` (sad trombone sound again).  Still, if we could rely on being able to use either `nil?` or `empty?` that would be nice.

I don't want to make too much of an issue of this at the moment.  I just had two small observations:  I had a momentary distraction of tracking down `from_syn(match)` to see if it was right, and I realized that I wasn't sure about how we are representing "empty" or "false" things.

no-reply:
@markbreedlove I've sometimes been tempted to introduce a `NullObject` pattern to deal with cases like this.

See, e.g. http://devblog.avdi.org/2011/05/30/null-objects-and-falsiness/

I'm going to float a solution that simplifies this slightly.

For the `[]=` return, I added a comment that might help a bit.  I propose to leave it like that.  I think this comes down to a matter of language conventions.

no-reply:
I ended up removing the `nil` return values in favor of returning the node-y  language `Resource`.  `#enrich_literal` now handles rejecting bnodes as `skos:exactMatch`.  I think this is a bit more explicit, in addition to avoiding the `nil` returns.

Ready to :squash:?

anarchivist:
:+1:

no-reply:
:rocket:

PR #163, opened 2015-05-28
Add boolean or expressions to Krikri::Parser::Value subclasses

markbreedlove:
Add the ability to interpret boolean "or" expressions in the names passed to `Krikri::XmlParser#get_child_nodes` and `Krikri::JsonParser#get_child_nodes`, by way of `Krikri::Parser::ValueArray#field` and `#fields`.

This is necessary so that we can do the following sort of thing in our mappings:
```
    providedLabel record.field('description', 'item|itemAv|fileUnit',
                               'physicalOccurrenceArray',
                               'itemPhysicalOccurrence|' \
                                 'itemAvPhysicalOccurrence|' \
                                 'fileUnitPhysicalOccurrence',
                               'referenceUnitArray', 'referenceUnit',
                               'name')
```
Both `XmlParser` and `JsonParser` have been modified so that usage is consistent between the two.

### DIFF
--- a/lib/krikri/parsers/json_parser.rb
+++ b/lib/krikri/parsers/json_parser.rb
@@ -44,6 +44,11 @@ module Krikri
 
       private
 
+      ##
+      # @see Krikri::Parser#get_child_nodes
+      #
+      # @param name_exp [String]  Object property name
+      # @return [Krikri::Parser::ValueArray]
       def get_child_nodes(name)
         if @node[name].is_a?(Array)
           vals = @node[name].map { |node| self.class.new(node) }
--- a/lib/krikri/parsers/xml_parser.rb
+++ b/lib/krikri/parsers/xml_parser.rb
@@ -61,9 +61,16 @@ module Krikri
 
       private
 
+      ##
+      # @see Krikri::Parser#get_child_nodes
+      #
+      # @param name_exp [String]  Element name
+      # @return [Krikri::Parser::ValueArray]
       def get_child_nodes(name)
-        Krikri::Parser::ValueArray.new(@node.xpath("#{@node.path}/#{name}", @namespaces)
-          .map { |node| self.class.new(node, @namespaces) })
+        Krikri::Parser::ValueArray.new(
+          @node.xpath("#{@node.path}/#{name}", @namespaces)
+            .map { |node| self.class.new(node, @namespaces) }
+        )
       end
 
       def attribute(name)
--- a/lib/krikri/parser.rb
+++ b/lib/krikri/parser.rb
@@ -36,12 +36,28 @@ module Krikri
     # child nodes and attributes.
     class Value
       ##
-      # Property accessor interface. Passes `name` to the local implementation
-      # of #get_child_nodes.
+      # Property accessor interface. Passes a name expression (`name_exp`) to
+      # the local implementation of #get_child_nodes.
       #
-      # @param name [#to_sym] a named property to access
-      def [](name)
-        get_child_nodes(name)
+      # The given name expression must follow the pattern:
+      #     name [| name ...]
+      #
+      # The optional "|" is a short-circuit operator that will return the
+      # property or element in the document for the first matching part of the
+      # phrase.
+      #
+      # @example
+      #   va = value['title']
+      #   va = value['this|that']  # gives value for "this" if both defined
+      #
+      # @param name [String] An expression of named properties to access
+      # @return [Krikri::Parser::ValueArray]
+      def [](name_exp)
+        name_exp.strip.split(/\s*\|\s*/).each do |n|
+          result = get_child_nodes(n)
+          return result unless result.empty?
+        end
+        Krikri::Parser::ValueArray.new([])
       end
 
       ##
@@ -112,9 +128,10 @@ module Krikri
       end
 
       ##
-      # @abstract Provide an accessor for properties
-      # @param name [#to_sym] a named property to access
-      # @return [Krikri::Parser::Value] the value of the child node
+      # @abstract Return a Krikri::Parser::ValueArray of child nodes
+      #
+      # @param name [String] Element or property name
+      # @return [Krikri::Parser::ValueArray] The child nodes
       def get_child_nodes(name)
         raise NotImplementedError, "Can't access property #{name}"
       end
--- a/spec/lib/krikri/parsers/json_parser_spec.rb
+++ b/spec/lib/krikri/parsers/json_parser_spec.rb
@@ -27,4 +27,13 @@ describe Krikri::JsonParser::Value do
   let(:record) { build(:json_record) }
 
   it_behaves_like 'a parser value'
+
+  it 'allows boolean "or" with "|" in the field name' do
+    expect(subject['notdefined|subject'].values).to eq ["Moomin Papa"]
+  end
+
+  it 'returns the value from the first defined name, given a "|"' do
+    # "subject" comes last in the document, but first in the "|" expression:
+    expect(subject['subject|title'].values).to eq ["Moomin Papa"]
+  end
 end
--- a/spec/lib/krikri/parsers/xml_parser_spec.rb
+++ b/spec/lib/krikri/parsers/xml_parser_spec.rb
@@ -19,4 +19,19 @@ describe Krikri::XmlParser::Value do
 
   it_behaves_like 'a parser value'
   it_behaves_like 'a parser value that has attributes'
+
+  context 'with a root path' do
+    subject { Krikri::XmlParser.new(record, '//oai_dc:dc').root }
+
+    it 'allows boolean "or" with "|" in the field name' do
+      expect(subject['dc:notdefined|dc:type'].values).to eq ["model"]
+    end
+
+    it 'returns the value from the first defined name, given a "|"' do
+      # "dc:type" comes last in the document, but comes first in the "|"
+      # expression:
+      expect(subject['dc:type|dc:title'].values)
+        .to eq ["model"]
+    end
+  end
 end


markbreedlove:
@no-reply and @anarchivist:

There has been some discussion outside of here about the `|` token.   I'd be happy to rename it to "or" to avoid confusion with XPATH, because it's not documented as producing unions.  The need eventually for unions is acknowledged, but something to handle with another operator; out of scope for the time being, since we're biting off a smaller piece now, just trying to get NARA mapped.

It's documented as being a boolean "or", and I intended the operator to work like "or" in Perl or Python( `0 or 'a'` resulting in 'a' in either, `[] or ['a']` resulting in `['a']` in Python).

It is also supposed to be bound more loosely than anything else, and the functionality could be expanded with other operators later.

Regarding an issue raised about the order of the elements in the provider's document being potentially different, and the way I'm grabbing the first defined element in the document:  I think both fields existing at the same time would constitute ambiguity in the metadata that would be best worked out with the provider.  I think that we might not want to make any promises about which field gets taken first.  Supposing it was something we wanted to accept, it's not easy to work with given way we're currently using XPATH in `#get_child_nodes`.  Since we don't want to pass through XPATH expressions from the DSL to the XML parser, why use XPATH at all in `#get_child_nodes`?  Why not use some easier way that lets us refer to the document as a hash?

I'm writing this quickly and have to resume this train of thought later today ...

markbreedlove:
Just another thought while I have a second:

If we have a provider with multiple elements defined for the same thing, why not use a union operator to get them both and then evaluate them in the enrichment phase?  It seems that we shouldn't care which one is preferred in the "or" because that should be reserved only for cases like NARA where we know it's really supposed to be one or the other.

We want to make sure we don't make enrichment decisions in the mapping phase or introduce too much code to the parser classes.  We just want to grab one _or_ the other element, or grab a union of elements and sort it out in enrichment.

no-reply:
Thanks @mb.

I'm in favor of taking on DSL extension with the limited scope needed to accomplish the NARA mapping, as you suggest.

My main stumbling block the use of the terminology "boolean or" for a short-circuited, sequence point/control flow or.  I *think* I'm okay with "or"... I find it somewhat counter-intuitive

> I think both fields existing at the same time would constitute ambiguity in the metadata that would be best worked out with the provider.

I agree that this is the case for the NARA mapping, but I'm not sure it holds for other uses of a sequence point selector.  It seems reasonable to use such an operator to do something like `record.field('dc:title|rdf:label')` expecting to use label as a fall-back when title is present. I don't think we need to resolve this now, necessarily, but I'm leary of _specifying_ the operator with this ambiguity built in.

On a related note, I'm unsure that a sequence point operator correctly specifies that mapping for NARA. My understanding is that we want to map `description.itemAV` if `record_type` matches some value, right?  Could/should we just implement that conditional?

> Since we don't want to pass through XPATH expressions from the DSL to the XML parser, why use XPATH at all in #get_child_nodes? Why not use some easier way that lets us refer to the document as a hash?

Agreed. I don't see us as pinned to XPATH at all, here. The various parsers can use whatever tools are at their disposal to implement the parser-layer.

We should probably just schedule a quick meeting to talk this out. I don't want to fuss too much about getting this perfect; I'll be happy if we can just get on the same page. :)

anarchivist:
@markbreedlove said, across two comments (emphasis added)...

> Regarding an issue raised about the order of the elements in the provider's document being potentially different, and the way I'm grabbing the first defined element in the document: **I think both fields existing at the same time would constitute ambiguity in the metadata that would be best worked out with the provider.**

...and...

>  It seems that we shouldn't care which one is preferred in the "or" because that *should be reserved only for cases like NARA where we know it's really supposed to be one or the other*.

I think this gets at the crux of this issue, at least for the issue what the key within the top-level `description` object in the NARA data is, and the subsequent issues. In my understanding there shouldn't be that ambiguity. Hence, this is why my inclination was something like "wildcard" operator would be useful here - because we don't care which element - we just care that something matches. 

@no-reply wrote:

> On a related note, I'm unsure that a sequence point operator correctly specifies that mapping for NARA. My understanding is that we want to map `description.itemAV` if `record_type` matches some value, right? Could/should we just implement that conditional?

I'm not sure I follow. What is `record_type`? My understanding is that we're not *too* choosy about the type we receive, but that sounds like a mapping question. The major issue is that the key/element under `description` determines elements further down the chain. We see that with other parts of the crosswalk too, e.g. the structures for creator and contributor.

no-reply:
@anarchivist the "record type" thing was related to me here:  https://dpla.slack.com/archives/tech/p1432310383000203

I'm not sure what value `record_type` would be in NARA's metadata, but my understanding is that it would be an error to collect `description.item` for items of a certain types, even if it were present. (I understand that NARA's API isn't supposed to give  us a `description.item` in those cases).

> Hence, this is why my inclination was something like "wildcard" operator would be useful here - because we don't care which element - we just care that something matches.

It's not clear to me whether your "wildcard" operator works like @markbreedlove's sequence point operator, or like union, or neither.  If I do `record.field('description', wild, wild,  'referenceUnitArray', ...)` am I selecting all nodes underneath `'description'`, just something that matches, or something else I'm overlooking?

----

I'm realizing that the main disconnect here is that we dealing with two different models, and seeing each this operator working in different contexts. I'm reading `record.field('title')` as an element selector; `record.field('title|label')`, for me, operates on that level, and `|` is an operator that works over the sets selected by 'title' and 'label'. This leads me to read "boolean or" as union.

The other model at play is control flow--"what do I do if the first field returns empty"?  The implementation @mb is arguing for is a parallel control flow or; i.e. "evaluate these expressions; give me one that selects non-empty set, I don't care which." Another approach would be a sequenced control flow or; "evaluate these expressions in order; give me the first that selects a non-empty set."

As I puzzle this out, all of these options are growing on me... Whatever we end up doing, I think we should avoid calling it "boolean or". :)

Talking this out with a language design chat yielded a couple of suggestions:

  - "you might also consider + for union, | for sequencing"
  - " in my work language i called that "rescue" bc missing values propagate like exceptions. any function of a "missing" expr is also "missing" and the | syntax "rescues" it"
  - ""defaults" might be a better word in that case too; default as in removing the fault of missing data"

markbreedlove:
The "or" I mean is disjunctive, in that it expresses a choice between two exclusive possibilities; you get one or the other, but not both.

I haven't heard of a "sequence point selector" before, and it's not a common term, but I am familiar with the notion of a short-circuit evaluation: http://en.wikipedia.org/wiki/Short-circuit_evaluation.  I am pretty confident in my usage of "boolean or" to mean a "boolean operator" or "logical operator"; the one for "or" -- which is "||" in many languages, or "or" (where the difference is often just precedence).

I chose "|" because it's more compact, and is used in Perl-compatible regular expressions (where the conditions under which you get a union from using it have to do with it being employed with a character class ("[]") or a regex modifier like "g".

I wasn't thinking of this originally, but it turns out that the operator in the regex stops the evaluation with the first thing it encounters _in the string_ that matches one of the conditions: try `"abc".match /(c|b|a)/` in Ruby, or the equivalent in any language that uses libpcre and you'll get "a", not a union of "c", "b", and "a".  Not that this has to operate exactly like a regex, but nobody should be looking at it, thinking, "I'm going to get a union from this."

I originally intended the operator to have short-circuited evaluation like "or" or "||", where it's evaluated from left to right, but then ended up with an XPATH issue where I couldn't pull that off conveniently.  All the same, it believe that its behavior will not surprise anyone, especially if it's documented, which it will be.

> My understanding is that we want to map description.itemAV if record_type matches some value, right? Could/should we just implement that conditional?

I'm trying to alleviate extra typing and extra lines of code in the mappings, and trying to make things easy for the authors of the mappings, surprising them as little as possible.  Using "|" the way it works now (or "or" if that's better) is going to be easier for a non-programmer to learn and type than using extra "if" statements.  I don't think that most of the mapping authors are going to be surprised by the way it works currently.

> The other model at play is control flow--"what do I do if the first field returns empty"? The implementation @markbreedlove is arguing for is a parallel control flow or; i.e. "evaluate these expressions; give me one that selects non-empty set, I don't care which."

That's not what it's doing; it's explicitly documented as having a particular order; it happens to be what appears first in the document.  It doesn't have to be exactly like regular expressions or XPATH or anything else; it has to be well-documented and convenient for the problem it's there to solve.

I agree with putting in a "+" or "*" operator at some future date to get unions.

no-reply:
Okay, I'm going to step aside on this one. :)

I think the syntax and behavior here are okay and I don't want to hold this up with circular discussions.  Sorry this was a derail.

lib/krikri/parsers/xml_parser.rb
```diff
@@ -61,9 +61,81 @@ def values?
...
+        # An assumption is made that no parser is really going to be used in
+        # a mapping without a root path (as defined by the root_path
+        # parameter to its constructor).  The usual XPATH expression will
+        # tend to be the first one.  The predicate will not work with
+        # @node.path == '/*', which is the default when the root path is not
+        # specified.
+        #
+        # When the root path is specified, name values with the "xmlns" prefix
+        # will not work; since this is the default namespace, we want to refer
+        # to them as the un-prefixed name.  For example, inside an OAI-PMH
```
anarchivist: Hm, this feels strange to me, since the XPath shouldn't match unless `xmlns:` is specified (see [here](http://tenderlovemaking.com/2009/04/23/namespaces-in-xml.html) for example). Could you add a spec that tests this with something from the OAI-PMH schema?

anarchivist: Yeah, this definitely needs a spec, because `xmlns:` elements are currently resolving appropriately, e.g. in the [ESDN mapping](https://github.com/dpla/heidrun-mappings/blob/master/heidrun/esdn_mods.rb).

anarchivist:
I think this generally looks OK, but as noted above, I want to verify we're not introducing a breaking change for elements in the default namespace (i.e. `xmlns:`).

lib/krikri/parsers/json_parser.rb
```diff
@@ -44,7 +44,32 @@ def values?
...
+      # or "name expression" of exclusive names separated by "|" characters.
+      #
+      # The values come from the property that comes first in the JSON object,
+      # which is given in the name expression.  This behavior maintains
+      # consistency with Krikri::XmlParser#get_child_nodes, which takes the
+      # first value in the document due to way in which XPATH is employed.
+      #
+      # @see Krikri::XmlParser#get_child_nodes
+      #
+      # @example
```
no-reply: I think it would be a good idea to revisit some of this documentation. It seems like some of it should sit in the base `Krikri::Parser` abstract method. The specific implementations should point back to that, and add any specific details like " The values come from the property that comes first in the JSON object which is given in the name expression...".

lib/krikri/parsers/xml_parser.rb
markbreedlove: My doc comment was not worded very well.

The ESDN mapping, it's using Krikri::ModsParser, which is instantiated with [`root_path = '//mods:mods'`](https://github.com/dpla/KriKri/blob/v0.5.7/lib/krikri/parsers/mods_parser.rb#L9).  That's good -- it will cause `#xpath_expression` to get called where `@node.path` is some long path with a lot of parts that includes 'mods:mods' ... I don't have an example handy.  It won't be '/*'

The string "xmlns:setSpec" (for example) as a field name would result in a "hit" before this revision with `"#{@node.path}/#{name}"`.  However, "(#{@node.path}/*[name() = 'xmlns:setSpec'])[1]" does not work.  But "(#{@node.path}/*[name() = 'setSpec'])[1]" *does* work.  The `name()` function returns literally the `name` property of the element, not the prefix plus the name.  You end up testing whether 'setSpec' is equal to 'xmlns:setSpec', which is false.

For this reason, I have to strip out 'xmlns:' in order to construct this XPATH predicate.

Can you actually help me understand why "xmlns:" would have to be used to prefix the `setSpec` element name?  If `setSpec` exists in the default namespace of the document, why can't we refer to it as `setSpec`?

no-reply: > Can you actually help me understand why "xmlns:" would have to be used to prefix the setSpec element name? If setSpec exists in the default namespace of the document, why can't we refer to it as setSpec?

I think this is one of those really hard to follow things about XML namespaces.  I'll try tell it how I understand it, with the caveat that I sometimes get turned around on this...

When you do `name()` you're asking for the QName of the tag __in context__ of the node's position in the document.  So a node named `http://example.org/my_tag` gets squashed to `ex:my_tag` or `my_tag`, depending on the namespace declarations that apply to the context.

An XPath expression, by contrast, is independent of the document context.  The namespaces applied are the ones declared as in scope for the _expression_.  This way a given expression finds semantically equivalent nodes across documents with different namespaces or namespace scoping.  
In XPath 1.0 the expression context is said to include ["the set of namespace declarations in scope for the expression"](http://www.w3.org/TR/xpath/#section-Introduction).  There's no default namespace here, just a convention to use "xmlns" when you know you're querying the default namespace in the document/element context.

In 2.0, namespace declarations are part of the "Static Context" of the expression (see [XPath 2.0 sec. 2.1.1](http://www.w3.org/TR/xpath20/#static_context)), and there is the possibility for a default namespace on the expression.

I'm not sure what any of this means for these changes, but I'm guessing there's a way to tighten this up a bit.  One issue that comes to mind is that using `name()` may be problematic if someone's XML has different namespace declarations/scoping across records.

no-reply: Looking at this closer, I'm wondering if the best approach isn't to do something like this:

In `Krikri::Parser`:

```ruby
def [](name)
  name.split('|').each { |n| result = get_child_nodes(n); return result unless result.empty? }
end
```

In this case: 
  - the `#get_child_node` implementations stay the same, with a simplified requirement for new parser implementations.
  - we can document that `Krikri::Parser::Value#[]` handles all internal field syntaxes like `|`, so we shouldn't see bits behavior for parsing those syntaxes spreading around multiple implementation modules.
  - we can revert to the original idea that we accept the first matched field from the user input.
  - there may be a minor performance cost since we're running multiple (e.g.) XPath queries through `#get_child_node` calls. But it seems like a cost worth bearing for the shaper interface.

Am I missing something here? Thoughts?

----

More realistically, this implementation fails only the tests that explicitly test "first defined in document" behavior. It probably needs some tweaking to deal with edge cases, and `result = nil` smells bad to me (we could end up returning `nil`). But proof-of-concept:

```ruby
def [](name)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
  result = nil                                                                                                                                                                                                                                                          
  name.split('|').each do |n|                                                                                                                                                                                                                                           
    result = get_child_nodes(n)                                                                                                                                                                                                                                         
    return result unless result.empty?                                                                                                                                                                                                                                  
  end                                                                                                                                                                                                                                                                   
  result                                                                                                                                                                                                                                                                
end 
```

lib/krikri/parsers/json_parser.rb
```diff
@@ -44,7 +44,17 @@ def values?
...
       private
 
+      ##
+      # @see Krikri::Parser#get_child_nodes
+      #
+      # @param name_exp [String]  Object property name
+      # @return [Krikri::Parser::ValueArray]
       def get_child_nodes(name)
+        return child_node(name) if @node.include?(name)
+        Krikri::Parser::ValueArray.new([])  # nothing found
```
no-reply: Can you say something quickly about why this change is necessary? Does `Krikri::Parser::ValueArray.new(vals)` not come up with `Krikri::Parser::ValueArray.new([])` when nothing is found, below?

markbreedlove:
https://github.com/dpla/KriKri/commit/2d14c17194f3d3fa2fa1645d12bfb50af0667b2e changes the `|` handling again to do what @no-reply suggested with moving the `|` up to `Krikri::Parser::Value#[]`.   This avoids redundancy between the JSON and XML parsers and changes the order of operation back to the way it was originally intended.

no-reply:
:rocket: this looks good to me.

I had a quick question about an implementation detail on `JsonParser`, but it's very minor and I'm probably just not following the program's logic properly.

I'm :+1: to merge; doubly so if my confusion can be cleared up. :)

lib/krikri/parsers/json_parser.rb
markbreedlove: Could just be an oversight after refactoring.  I'll check in an hour or so.

no-reply:
:100: 

@anarchivist Are you :+1: to :squash:

anarchivist:
:ok_hand:

markbreedlove:
Squashed, and CI is running ...

no-reply:
Thanks @markbreedlove!

markbreedlove:
Sure thing, thanks for your help!

PR #164, opened 2015-05-29
Extract Enrichment/FieldEnrichment to Audumbla

no-reply:
This replaces `Krikri::Enrichment` and `Krikri::FieldEnrichment` with `Audumbla`. This is intended to allow enrichment processes to grow outside of the context of KriKri, for generic enrichments to be shared as part of `Audumbla`, and for enrichments to develop as stand-alone code.

`Audumbla` currently contains the main `Enrichment` and `FieldEnrichment` modules and shared examples for RSpec.  We should continue to extract some of the specific enrichments that seem of general use.

As this develops, I'd like to move *all* enrichments out of `Krikri`, either up into `Audumbla` or down into `Heidrun`.

~~`audumbla` 0.1.0 is currently failing CI due to an issue with the FactoryGirl linter.  Wait to merge this until that is resolved--there's a slim chance we'll want to push a patch release before pinning `Krikri` to the gem.~~


anarchivist:
> As this develops, I'd like to move all enrichments out of `Krikri`, either up into `Audumbla` or down into `Heidrun`.

:+1:

anarchivist:
I'm good with this. Someone else want to take a crack at reviewing? @markbreedlove @AudreyAltman

markbreedlove:
This makes sense.

At what point do you think the enrichments should be moved?  Is there anything keeping them from being moved now?

no-reply:
> Is there anything keeping them from being moved now?

Nothing except making decisions about where individual enrichments should live & taking the time to move them.

The namespaces will change, so some of our "Enrichment Profile" examples will need to be kept in sync.

If we want to hold off on this until some or all of them have been moved, I'm okay with that. This will be a blocker for the Coarse Geocoder within the next couple of days--so that's a factor.

markbreedlove:
I don't have any problem with merging this in the interest of time, and coming back to it later to move all of the enrichments to `audumbla`.

AudreyAltman:
I generally understand and support the idea of moving enrichments to their own repo.  I don't think I'm familiar enough with this part of the software to anticipate any specific potential problem areas.  The code itself looks good to me.  Please let me know if there's any localized testing I can do to help.

PR #166, opened 2015-06-08
Do not obliterate defaults, QASearchIndex constructor

markbreedlove:
Have the `QASearchIndex` constructor preserve default configuration
properties that are not being passed in via the `opts` argument to
`#initialize`.

This allows the Solr URI in Krikri::Settings.solr to be passed to
`RSolr.connect`.

Fixes https://www.pivotaltracker.com/story/show/91822774

### DIFF
--- a/lib/krikri/search_index.rb
+++ b/lib/krikri/search_index.rb
@@ -136,8 +136,10 @@ module Krikri
     ##
     # @param opts [Hash] options to pass to RSolr
     # @see RSolr.connect
-    def initialize(opts = Krikri::Settings.solr.to_h)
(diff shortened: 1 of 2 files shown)


no-reply:
That would do it!  Do we care to get a regression test in for this?  Seems like it would be a good idea.

markbreedlove:
@no-reply: OK, I've added a test.

no-reply:
Awesome. Looks ready to me. :squash:

anarchivist:
I'm also :+1:.

markbreedlove:
@anarchivist @no-reply: squashed

PR #168, opened 2015-06-22
Fix broken build on enrichments

anarchivist:
The order of the last few merges ended up breaking the build. In the process of resolving this, I discovered that RSpec 3.3 also breaks the build, but I think that will have to wait for @no-reply to look at.

There's a remaining outstanding issue where we're seeing the following error:

```
undefined method `type' for .focus:Sass::Selector::Class (NoMethodError)
```

...which was first discovered by @markbreedlove.


anarchivist:
Closing; see PR #169 instead.

PR #169, opened 2015-06-24
Fix broken builds from enrichment merges and RSpec/Sass issues

anarchivist:
This addresses the following issues:

* The order of the last few merges ended up breaking the build (when we merged the Audumbla abstraction before merging the language-to-Lexvo enrichment).
* Since the last release of Krikri/Heidrun, RSpec 3.3 was released, and running the specs under it breaks the build. I'm not sure why but I think that will have to wait for @no-reply to look at, because I'm guessing it's an issue with `rdf-spec` or `active_triples`. 
* This also addresses the `undefined method 'type' for .focus:Sass::Selector::Class (NoMethodError)` failure when precompiling the assets or running the specs for the `HarvestSourcesController`. The changes feel a bit excessive, but the next most feasible workaround would be to upgrade Krikri to Rails 4.2.

Closes #168.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,6 +35,6 @@ else
     gem 'sass-rails', '~> 5.0.0'
     gem 'responders', '~> 2.0'
   else
-    gem 'sass-rails', '~> 4.0.3'
+    gem 'sass-rails', '~> 5.0.0'
   end
 end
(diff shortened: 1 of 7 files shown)


anarchivist:
Superseded by #170.

PR #171, opened 2015-06-29
Handle empty/non-existent root nodes on XmlParser

no-reply:
If a non-existent path is given as the root node for a record, parsing failed with `NoMethodError` on an `#xml?` call to `nil`. This makes the parser throw a sensible error instead.

This should take care of error handling for the current ESDN mapping. It's an incremental improvement on error handling in mapping in general.

While we're in here, I'm open to suggestions about what to output as error context when `rec` is `nil`. `e.message` and `e.backtrace` may be enough.

### DIFF
--- a/lib/krikri/mapper.rb
+++ b/lib/krikri/mapper.rb
@@ -135,7 +135,7 @@ module Krikri
             rec.mint_id! if rec.node?
             activity_uri ? rec.save_with_provenance(activity_uri) : rec.save
           rescue => e
-            Rails.logger.error("Error saving record: #{rec.rdf_subject}\n" \
+            Rails.logger.error("Error saving record: #{rec.try(:rdf_subject)}\n" \
                                "#{e.message}\n#{e.backtrace}")
           end
         end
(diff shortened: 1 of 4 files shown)


lib/krikri/mapper.rb
```diff
@@ -135,7 +135,7 @@ def run(activity_uri = nil)
             rec.mint_id! if rec.node?
             activity_uri ? rec.save_with_provenance(activity_uri) : rec.save
           rescue => e
-            Rails.logger.error("Error saving record: #{rec.rdf_subject}\n" \
+            Rails.logger.error("Error saving record: #{rec.try(:rdf_subject)}\n" \
```
markbreedlove: That's interesting.  I'm not familiar with `try` as a method.  Can you tell me where it's defined, and what classes inherit it?  (Where could I expect to be able to use it?)

markbreedlove:
Looks good. 

As for what to log:  If it has an rdf_subject, it allows us to find the original record to retrieve the XML.   I don't how to improve things, other than eventually adding tagged logging, which is the subject of another story.  More improvements should come later when we think about error-handling in more depth, and storing error-related statistics with the activity.

lib/krikri/mapper.rb
no-reply: `#try` is a method on `Object`.  It's basically `#send` except it returns `nil` instead of throwing an error if the method is undefined.

http://apidock.com/rails/Object/try

PR #172, opened 2015-06-30
Feature/optimize ldp

no-reply:
This removes post-save HEAD requests from harvests.


markbreedlove:
Looks good, and I'm eager to see how this works.

no-reply:
Do you think I should :squash:, or do you like these as separate commits?

markbreedlove:
I think they should be squashed, because `save` was modified in the first commit, and the addition of `save_and_reload` goes along with it.

PR #173, opened 2015-07-06
Log requests on OAI Harvests

no-reply:
Sets up the Faraday client that makes OAI requests to log requests; this makes it possible to spot retried requests.

### DIFF
--- a/lib/krikri/harvesters/oai_harvester.rb
+++ b/lib/krikri/harvesters/oai_harvester.rb
@@ -42,6 +42,7 @@ module Krikri::Harvesters
       http_conn = Faraday.new do |conn|
         conn.request :retry, :max => 3
         conn.response :follow_redirects, :limit => 5
+        conn.response :logger, Rails.logger
         conn.adapter :net_http
       end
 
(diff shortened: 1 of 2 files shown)


markbreedlove:
Does this log the message with a timestamp?  The reason I ask is because we added tagged logging for the SoftwareAgent, and mean to elsewhere, partly because we were seeing messages without timestamps.

no-reply:
No, this should just log what the default Faraday logger does: https://github.com/lostisland/faraday/blob/81f16593a0138ec58bb6f25e1c2804e91589662f/lib/faraday/response/logger.rb

I'm open to suggestions about other ways to handle this.

markbreedlove:
In the console, if I create a new Faraday object with the logger set to stdout, and pass that to the OAI harvester, the output it produces has timestamps.  Would this be the same thing that would be logged if I used `Rails.logger`?  If so, it should be good for now.

```
[1] pry(main)> http_conn = Faraday.new do |conn|
[1] pry(main)*   conn.response :logger
[1] pry(main)*   conn.adapter :net_http
[1] pry(main)* end  
[2] pry(main)>  client = OAI::Client.new('http://localhost/', :http => http_conn)
[5] pry(main)> client.list_sets
I, [2015-07-07T13:49:12.994986 #10662]  INFO -- : get http://localhost/?verb=ListSets
D, [2015-07-07T13:49:12.995408 #10662] DEBUG -- request: User-Agent: "Faraday v0.9.1"
[ ... ]
```

markbreedlove:
I'm fine with this if nobody else has more observations.

PR #174, opened 2015-07-07
Adds solr commit to solr bulk add processes

no-reply:
The QA index perviously required a manual commit following indexing jobs. `#update_from_activity` continues to return the same, but calls `solr#commit` after running its bulk adds.


markbreedlove:
That looks good to me.

no-reply:
:cookie: Great. Does @AudreyAltman or @anarchivist want to take a look?

AudreyAltman:
:+1:

PR #175, opened 2015-07-07
Adds `#reject_attribute` to the DSL `ValueArray`

no-reply:
Expands the functionality of `#match_attribute` to accept a block to match against and adds an inverse `#reject_attribute` that deals with some cases brought on by Tennessee's MODS.

### DIFF
--- a/lib/krikri/parser.rb
+++ b/lib/krikri/parser.rb
@@ -273,17 +273,54 @@ module Krikri
       end
 
       ##
+      # @example selecting by presence of an attribute; returns all nodes where
+      #   `#attribute?(:type)` is true
+      #
+      #   match_attribute(:type) 
+      #
+      # @example selecting by the value of an attribute; returns all nodes with
+      #   `#attribute(:type) == other`
+      #
+      #   match_attribute(:type, other)
+      #
+      # @example selecting by block against an attribute; returns all nodes with
+      #   `block.call(attribute(:type))` is true
+      #
+      #   match_attribute(:type) { |value| value.starts_with? 'blah' }
+      #
+      # @example selecting by block against an attribute; returns all nodes with
+      #   `block.call(attribute(:type)) == other` is true
+      #
+      #   match_attribute(:type, 'moomin') { |value| value.downcase }
+      #   
       # @param name [#to_sym] an attribute name
-      # @param other [Object] an object to for equality with the
+      # @param other [Object] an object to check for equality with the
       #   values from the given attribute.
       #
+      # @yield [value] yields each value with the attribute in name to the block
+      #
       # @return [ValueArray] an array containing nodes for which the specified
-      #   attribute has a value matching the given object.
-      def match_attribute(name, other)
-        select do |v|
-          next unless v.attribute?(name.to_sym)
-          v.send(name).downcase == other.downcase
-        end
+      #   attribute has a value matching the given attribute name, object, and 
+      #   block.
+      def match_attribute(name, other = nil, &block)
+        select(&compare_to_attribute(name, other, &block))
+      end
+
+      ##
+      # @param name [#to_sym] an attribute name
+      # @param other [Object] an object to check for equality with the
+      #   values from the given attribute.
+      #
+      # @yield [value] yields each value with the attribute in name to the block
+      #
+      # @return [ValueArray] an array containing nodes for which the specified
+      #   attribute does not have a value matching the given attribute name, 
+      #   object, and block.
+      #
+      # @see #match_attribute  for examples; this calls #reject, where it calls
+      #   #select.
+      def reject_attribute(name, other = nil, &block)
+        reject(&compare_to_attribute(name, other, &block))
       end
 
       ##
@@ -299,6 +336,20 @@ module Krikri
 
       def get_field(name)
         self.class.new(flat_map { |val| val[name] })
+      end
+
+      private
+      
+      ##
+      # @see #match_attribute, #reject_attribute
+      def compare_to_attribute(name, other, &block)
+        lambda do |v|
+          next unless v.attribute?(name.to_sym)
+          result = v.send(name)
+          result = yield(result) if block_given?
+          return result == other if other
+          result
+        end
       end
 
       public
--- a/spec/lib/krikri/parser_value_array_spec.rb
+++ b/spec/lib/krikri/parser_value_array_spec.rb
@@ -238,18 +238,61 @@ describe Krikri::Parser::ValueArray do
     end
   end
 
-  describe '#match_attribute' do
+  shared_context 'with attributes' do
     before do
       values.each { |val| allow(val).to receive(:attribute?).and_return(false) }
-    end
 
-    it 'selects values by their attributes' do
       allow(values[0]).to receive(:attribute?).with(:type).and_return(true)
       allow(values[0]).to receive(:type).and_return('Moomin')
       allow(values[1]).to receive(:attribute?).with(:type).and_return(true)
       allow(values[1]).to receive(:type).and_return('mummi')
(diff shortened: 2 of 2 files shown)


no-reply:
An example of the TN mapping:

```ruby
 relation record.field('mods:relatedItem')
  .reject { |rights| rights.try(:type) == 'isReferencedBy' }
  .reject { |rights| rights.try(:type) == 'references' }
  .field('mods:titleInfo')                                                                                                                                       
  .fields('mods:url', 'mods:title')
```

The above can now be done like:

```ruby
 relation record.field('mods:relatedItem')
  .reject_attribute(:type, 'isReferencedBy')
  .reject_attribute(:type, 'references')
  .field('mods:titleInfo')                                                                                                                                       
  .fields('mods:url', 'mods:title')
```

or better:

```ruby
 relation record.field('mods:relatedItem')
  .reject_attribute(:type) { |attr| attr == 'isReferencedBy' || attr == 'isReferencedBy' }
  .field('mods:titleInfo')                                                                                                                                       
  .fields('mods:url', 'mods:title')
```

anarchivist:
:+1:

markbreedlove:
Yes, this is more convenint.  Could I just check my understanding, as follows?

It appears that it's designed for XML with attributes, such as:
```
<some-parent-element>
    <mods:relatedItem type='isReferencedBy>something to reject</mods:relatedItem>
   <mods:relatedItem>
        <mods:titleInfo>
            <mods:url>http://example.tld/</mods:url>
           <mods:title>The title</mods:title>
        </mods:titleInfo>
    </mods:relatedItem>
</some-parent-element>
```

Should I understand `reject_attribute` and `match_attribute` to be XML-only methods?

no-reply:
> Should I understand reject_attribute and match_attribute to be XML-only methods?

I don't think so, no. These methods should work with any format that has nodes with attributes. The [`Krikri::Parser`](https://github.com/dpla/KriKri/blob/develop/lib/krikri/parser.rb#L97-L99), carries the relevant abstraction.

markbreedlove:
If I have JSON like this ...
```
{
  'relatedItem': {'type': 'isReferencedBy', '#text': 'something to reject'},
  'relatedItem': {'#text': 'good thing'}
}
```
... could I then write the following?
```
relation record.field('relatedItem')
  .reject_attribute(:type, 'isReferencedBy')
  .field('#text')
```

If that would work, then can you confirm that `match_attribute` or `reject_attribute` would also work if it were handed an object that's not a hash?  (If `record.field('relatedItem')` returned a string?)

It also may be out of scope, but I'm wondering what I'd do about an object that needs to be evaluated deeper than one level down, such as if the 'relatedItem' in the example above were `{'typeInfo': {'type': 'something'}}` and we wanted the whole 'relatedItem' where `typeInfo.type` is 'something'.  Is this taken care of somewhere?  (Would I just use a `reject` call in that case?)

no-reply:
> If I have JSON like this ...   ... could I then write the following?

No. Sorry, I think this discussion has been confusing.  The `Krikri::Parser` deals with nodes ("fields"), which *may* have attributes. The JSON parser considers `relatedItem`, `type`, `#text`, etc... to be nodes and does not support attributes.

Your example code will throw `NotImplementedError`* if run against a record parsed with the JSON parser---it will try to access attribute `type` and hit the unimplemented [`#attributes`](https://github.com/dpla/KriKri/blob/develop/lib/krikri/parsers/json_parser.rb#L29).

You *could* do something like:

```ruby
relation record.field('relatedItem')
  .reject { |n| n.child?(:type) }
  .field('#text')
```

> If that would work, then can you confirm that match_attribute or reject_attribute would also work if it were handed an object that's not a hash? (If record.field('relatedItem') returned a string?)

All operations run though `record` are eventually called on, and return a `Krikri::Parser::ValueArray`. That array may be empty from the start, or continue to have values to the end. This means that there's safe method-chaining as deep as you like. Eventually the `Declaration` (property or child) calls [Krikri::Parser::ValueArray#values](https://github.com/dpla/KriKri/blob/develop/lib/krikri/parser.rb#L188-L190) to get the, e.g. the String.

> It also may be out of scope, but I'm wondering what I'd do about an object that needs to be evaluated deeper than one level down, such as if the 'relatedItem' in the example above were {'typeInfo': {'type': 'something'}} and we wanted the whole 'relatedItem' where typeInfo.type is 'something'. Is this taken care of somewhere? (Would I just use a reject call in that case?)

It should be possible to do this, but it might not be as pretty as we'd like at this point.  I think something like `record record.field('relatedItem').select { |item| item[:infoType].field(:type).include? 'something'}` would work. This seems like a gap to me; I think a design goal should be for the user not to need access to anything lower on the parser tree than the current node when they use `#select` and `#reject`. Anything like this example really needs a DSL syntax.

Does that help? :D

*this is maybe not the best response, maybe parsers should be expected to explicitly have 0 attributes if they are unsupported? cc: @anarchivist

no-reply:
> This seems like a gap to me; I think a design goal should be for the user not to need access to anything lower on the parser tree than the current node when they use #select and #reject. Anything like this example really needs a DSL syntax.

To clarify this `record.field('relatedItem').select { |item| item.value == 'blah' }` seems like a good pattern. So does `record.field('relatedItem').map { |item| "somestuff ... #{item.value} "}. But `record.field('relatedItem').map { |item| item[:someField].field(:blah).map { |blah| blah.value }` should (and, in this case does) have a nicer syntax.

no-reply:
And here's a real life example of something I would like to abstract away: https://github.com/dpla/heidrun-mappings/blob/feature/TennMapping/heidrun/TN_mods.rb#L51

I think this line should be something more like:

`select_field('mods:role') { |role| role.include?('contributor') }`

anarchivist:
Discussion: 
> @markbreedlove:
>> Should I understand `reject_attribute` and `match_attribute` to be XML-only methods?
> @no-reply 
> I don't think so, no. These methods should work with any format that has nodes with attributes. The `Krikri::Parser`, carries the relevant abstraction.
> @markbreedlove:
>> If I have JSON like this ... ... could I then write the following?
> @no-reply:
>> No. Sorry, I think this discussion has been confusing. The Krikri::Parser deals with nodes ("fields"), which may have attributes. ...
>>
>> Your example code will throw NotImplementedError* if run against a record parsed with the JSON parser---it will try to access attribute type and hit the unimplemented #attributes. ... 
>>
>>*this is maybe not the best response, maybe parsers should be expected to explicitly have 0 attributes if they are unsupported? The more I think about this the more I'm convinced that it is very bad :tm: that attributes break method chaining for JSON records. cc: @anarchivist

Yeah, this was a decision in some ways that was made for the sake of expediency that I punted on while working on the `Krikri::JSONParser`. The `NotImplementedError is defined [here](https://github.com/dpla/KriKri/blob/develop/lib/krikri/parsers/json_parser.rb#L29-L31) ... and it's defined that way because, as the message says, `'Attributes are not supported for JSON'`.

Put somewhat differently, this is made most obvious by the distinction between JSON (its objects have keys, which *at the time* I thought were roughly like elements but get abstracted to "fields") and XML (which enforces a distinction between elements and attributes). I think part of the bias that's showing here is that the `Krikri::Parser` class we started with was the XML parser.

no-reply:
> I think part of the bias that's showing here is that the Krikri::Parser class we started with was the XML parser.

This is absolutely the source of a bias in the terminology for parsing. Suggestions for a better/more general terminology are welcome.

markbreedlove:
Thanks, @no-reply and @anarchivist.  That helps!

I don't have any objection to this being merged, if the topics being covered in this thread can be taken into consideration as the mapping DSL is developed further.  This changeset makes a commonly-occurring expression easier to write.

no-reply:
is this also ready to go? @markbreedlove: :merge: :grey_question:

PR #176, opened 2015-07-08
Use Ruby OAI's new `#_source` method to access XML

no-reply:
Rather than building a new XML document from scratch, it's now possible to use the REXML Element in the returned `OAI::Record` directly. This is the case from `ruby-oai` 0.4.0 on.

### DIFF
--- a/spec/lib/krikri/harvesters/oai_harvester_spec.rb
+++ b/spec/lib/krikri/harvesters/oai_harvester_spec.rb
@@ -188,6 +188,20 @@ EOM
                    :headers => {})
     end
 
+    it 'produces valid xml' do
+      expect do
+        Nokogiri::XML(subject.records.first.content) do |config|
+          config.options = Nokogiri::XML::ParseOptions::STRICT
+        end
+      end.not_to raise_error
+    end
+
+    it 'produces has oai namespace and header' do
+      expect(Nokogiri::XML(subject.records.first.content)
+              .xpath('//xmlns:header'))
+        .not_to be_empty
+    end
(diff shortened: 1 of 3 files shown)


no-reply:
Hold this. There may be an issue with the `xmlns` namespace

no-reply:
patched.

anarchivist:
:+1: - anyone else want to take a look? @markbreedlove @AudreyAltman

spec/lib/krikri/harvesters/oai_harvester_spec.rb
```diff
@@ -403,25 +417,6 @@
       end
     end
```
markbreedlove: Could you comment on why the tests were moved out of the `#record_xml` describe block?

markbreedlove:
It looks much better than it did before.  I was just curious about the relocation of the tests, because I could potentially overlook their existence without them being in a block named for the method.

spec/lib/krikri/harvesters/oai_harvester_spec.rb
no-reply: A couple of things: 
  - the older examples were unit tests on the private method `#record_xml`. These consider that method an implementation detail, and instead focus on the output of `#records`.
  - the old version wanted an empty `OAI::Record` to work with; these use the same test objects as the rest of the suite.

markbreedlove:
I don't think I have any more questions, if this wants to be squashed.

no-reply:
:squash:'d

PR #177, opened 2015-07-09
Fix/provider model

AudreyAltman:
This changes the `Krikri::Provider` model so that it populates with data from a small number of calls to Solr, rather than a large number of calls to Marmotta.  It should reduce page load times in the QA interface, especially in views that need to generate a list of all providers.

`Krikri::Provider` is changed from a `DPLA::MAP::Agent` object (which is an `ActiveTriples` object) to an `ActiveModel` object.  `ActiveModel` is much like `ActiveRecord` - it plays nicely with `ActionPack`and includes some model features.  However, `ActiveModel` abstracts away database interactions, allowing us to populate it with data from other sources like Solr.  An `ActiveModel` object is constructed by including and extending certain modules, and by including key methods (ie. `persisted?`).  We might consider using `ActiveModel` elsewhere in the code - for example, `Krikri::ValidationReport` seems a good candidate to be made into an `ActiveModel` object.  For more info on `ActiveModel`, see http://api.rubyonrails.org/classes/ActiveModel/Model.html or http://railscasts.com/episodes/219-active-model  

If you want to interact with with a provider as a `DPLA::MAP::Agent`, you can still do so; `Krikri::Provider` has an instance method, `#agent`, that returns the provider as a `DPLA::MAP::Agent` object.  You can even get said `DPLA::MAP::Agent` object without hitting Solr at all, provided you know the provider's `:rdf_subject` (aka it's full URI identifier).  Most of `Krikri::Provider`'s attributes are lazy-loaded so that time and memory can be expended on things like network calls only when necessary.

Another feature `Krikri::Provider` is that the `#all` and `#find` methods each make only one call to Solr, and then instantiate instances of `Krikri:Provider` with provider's `:rdf_subject` _and_ `:name`.  Before this change, generating each `:name` required its own separate network call, so this should also reduce page load times.

This PR required many changes to the code (especially specs) that use `Krikri::Provider`.  The changes are of three basic types:
1. Creating a `DPLA::MAP::Agent` object if needed
2. Adjusting for new attribute accessors (for example, using `#name` instead of `#label` to get the provider's name
3. Eliminating unnecessary network calls (for example, refraining from using `#find` if you don't really need to search Solr)

Page load time can still be improved with changes to the current model of tracking `@current_provider`s in the user interface, but that is outside the scope of this PR.

### DIFF
--- a/app/models/krikri/provider.rb
+++ b/app/models/krikri/provider.rb
@@ -1,80 +1,173 @@
 module Krikri
   ##
-  # This models provider data from the {Krikri::Repository}.
+  # Provider is an ActiveModel object. It represents a provider that has been
+  # indexed in Solr.
   #
-  # @todo create an appropriate `ActiveTriples::PersistenceStrategy' for
-  #   handling providers as data from a repository alongside LDP data.
-  # @see ActiveTriples::Resource
-  class Provider < DPLA::MAP::Agent
-    configure :base_uri => Krikri::Settings.prov.provider_base
+  # This is a read-only object. It does not write new providers to Solr, nor
+  # does it update or delete providers.
+  #
+  # A DPLA::MAP::Agent object can be constructed from an instance of this
+  # ActiveModel object using the :agent method. The DPLA::MAP::Agent object
+  # interacts with Marmotta, rather than Solr.
+  #
+  #   @example Krikri::Provider.find(rdf_subject: 'http://blah.com/123').agent
+  #     => [DPLA::MAP::Agent]
+  #
+  class Provider
+    include ActiveModel::Validations
+    include ActiveModel::Conversion
+    extend ActiveModel::Naming
+
+    ##
+    # @!attribute :rdf_subject [String] the URI identifying the provider
+    #   @example: "http://blah.com/contributor/123"
+    #
+    # @!attribute :id [String] the stub id
+    #   @example: "123"
+    #
+    # @!attribute :name [String] the human-redable name of the provider
+    #
+    # @!attribute :agent [DPLA::MAP::Agent] an ActiveTriples representation of
+    # the provider, read from Marmotta.
+    attr_accessor :rdf_subject, :name
+    attr_reader :agent, :id
+
+    ##
+    # Initializes a Provider object.
+    #
+    # @param attributes [Hash]
+    # If the params Hash contains a valid value for :rdf_subject, the model can
+    # extrapolate all other attributes.
+    #
+    # @example
+    #   Krikri::Provider.new({ rdf_subject: 'http:://blah.com/contributor/123',
+    #                          name: 'Moomin Valley Historical Society' })
+    #
+    # @raise [NoMethodError] if the params Hash includes a key that does not
+    # match listed attr_accessors
+    #
+    # @return [<Krikri::Provider>]
+    def initialize(attributes = {})
+      attributes.each do |name, value|
+        send("#{name}=", value)
+      end
+    end
 
     ##
     # Gives a list of all providers, ignoring those with no URI (bnodes).
+    #
+    # Providers will be initialized with both an :rdf_subject and a :name,
+    # if both exist in the Solr index.
     #
     # @return [Array<Krikri::Provider>]
     def self.all
       query_params = { :rows => 0,
                        :id => '*:*',
-                       'facet.field' => 'provider_id' }
+                       :facet => true,
+                       'facet.pivot' => 'provider_id,provider_name' }
       response = Krikri::SolrResponseBuilder.new(query_params).response
-      response.facets.first.items.map do |item|
-        provider = new(item.value)
-        provider.node? ? nil : provider
+
+      response.facet_pivot['provider_id,provider_name'].map do |facet|
+        rdf_subject = facet['value']
+        name = facet['pivot'].present? ? facet['pivot'].first['value'] : nil
+        provider = new({ :rdf_subject => rdf_subject, :name => name })
+        provider.valid_rdf_subject? ? provider : nil
       end.compact
     end
 
     ##
-    # @param id [#to_s] the identifier (local name or URI) of the provider
-    #   to find
+    # Finds a provider in Solr that matches the given ID, ignoring bnodes.
+    #
+    # @param [String] :id or :rdf_subject
     #
     # @return [Krikri::Provider]
-    # @todo Use {ActiveTriples}/{RDF::Repository} to populate the object
     def self.find(id)
-      return nil if id.nil?
-      provider = new(id)
-      provider.reload
-      return nil if provider.empty?
-      provider
+      rdf_subject = id.start_with?(base_uri) ? id : base_uri + id
+      query_params = { :rows => 1,
+                       :q => rdf_subject,
+                       :qf => 'provider_id',
+                       :fl => 'provider_id provider_name' }
+      response = Krikri::SolrResponseBuilder.new(query_params).response
+      return nil unless response.docs.count > 0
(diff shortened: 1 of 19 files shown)


AudreyAltman:
I have tested the QA interfaces locally with a couple sample records - it would also be good to test this with a large dataset.

app/models/krikri/provider.rb
```diff
@@ -1,80 +1,173 @@
...
+  #   @example Krikri::Provider.new(rdf_subject: 'http://blah.com/123').agent
+  #     => [DPLA::MAP::Agent] 
+  #
+  class Provider
+    include ActiveModel::Validations
+    include ActiveModel::Conversion
+    extend ActiveModel::Naming
+
+    ##
+    # @attr :rdf_subject [String] the URI identifying the provider
```
no-reply: We might consider using the [`@!attribute` tag](http://www.rubydoc.info/gems/yard/file/docs/Tags.md#attribute). `@attr` is deprecated.

AudreyAltman: Will do.

app/models/krikri/provider.rb
```diff
@@ -1,80 +1,173 @@
...
-      node? ? nil : rdf_subject.to_s.gsub(base_uri.to_s, '')
+    # Gives the last path fragment for :rdf_subject in HTML escaped form,
+    # ignoring :rdf_subjects, ignoring bnodes.
+    #
+    # @example
+    #   Given: rdf_subject == 'http://my_domain/blah/0123'
+    #   local_name == '0123'
+    #
+    # @return [String]
+    def local_name
```
no-reply: I think we now have this logic in a couple of places. (see: https://github.com/dpla/KriKri/blob/b52c2c958c13b08b3182e9e125e05ad02fb870f0/app/helpers/krikri/application_helper.rb#L29). I'm wondering what the best way to consolidate might be...

app/models/krikri/validation_report.rb
```diff
@@ -63,9 +63,11 @@ def find(id)
 
     private
 
+    ##
+    # Get the full URI identifier for the provider
     def provider_uri
       return unless provider_id.present?
-      Krikri::Provider.new(provider_id).rdf_subject
+      Krikri::Provider.base_uri + provider_id
```
no-reply: :+1:

app/models/krikri/provider.rb
AudreyAltman: Yes - I am hoping that this will be fixed when we change the way that `current providers` are tracked in the new IA.

app/models/krikri/provider.rb
```diff
@@ -1,80 +1,173 @@
...
     ##
-    # @return [String] the name of the provider
-    def provider_name
-      label.first || providedLabel.first || id
+    # Creates DPLA::MAP::Agent object from the :rdf_subject, ignoring bnodes.
+    #
+    # @return [DPLA::MAP::Agent]
+    def initialize_agent
+      return nil unless valid_rdf_subject?
+      DPLA::MAP::Agent.new(rdf_subject)
```
no-reply: Does this need to set the existing values to support, e.g. https://github.com/dpla/KriKri/pull/177/files#diff-e4132b9b98dd599bbfea544cea4ce2e1R139 ? 

Maybe something like 
```ruby
DPLA::MAP::Agent.new(rdf_subject).tap { |agent| agent.label = name }
```

spec/support/shared_examples/active_model_lint.rb
```diff
@@ -0,0 +1,14 @@
+shared_examples 'ActiveModel' do
+  include ActiveModel::Lint::Tests
```
no-reply: :+1:

spec/models/provider_spec.rb
```diff
@@ -37,90 +46,113 @@ module Isomorphic
...
+    end
 
-      let(:provider) { DPLA::MAP::Agent.new }
+    context 'with bnode' do
+      include_context 'bnode indexed in Solr'
 
-      it 'ignores bnodes' do
-        expect(described_class.all).to be_empty
+      it 'ingnores bnode' do
+        expect(described_class.all.count).to eq 0
```
no-reply: Does this no longer return an array that responds to `#empty?` ?  It's still documented as doing so at https://github.com/dpla/KriKri/pull/177/files#diff-c351630f46375b8711be0693562042e8R63

no-reply:
This looks good to me. The most substantial questions I have are at https://github.com/dpla/KriKri/pull/177#discussion_r34579357 and https://github.com/dpla/KriKri/pull/177#discussion_r34579888.

FWIW, I tried swapping out the `ActiveModel` `include`s and `extend`s for a `DPLA::MAP::Agent` subclass ([AM is added by `ActiveTriples`](https://github.com/ActiveTriples/ActiveTriples/blob/develop/lib/active_triples/rdf_source.rb#L51-L65)), leaving all the solr-driven code in place. I hit an error surrounding `#name` on the view side (that the error only shows up in the view suggests there is a gap in the unit testing; I'll look closer at this later today). 

But this also suggests that we could aim for an indexing strategy that allows us to populate our `DPLA::MAP` models in full from the index. If we got that in place, we could reduce back to a single model for `Provider`.

app/models/krikri/provider.rb
AudreyAltman: I had originally left the name out here b/c I was trying to avoid unnecessary calls to Solr - if the `Krikri::Provider` object is not instantiate with a `name` param, the object will call Solr to get the name.  But now that I think about it, I think that constructing a better `agent` object is worth the cost of a possible call to Solr.  If you agree, I will go ahead and make your suggested change.

app/models/krikri/provider.rb
no-reply: Perfect. I'm happy to punt it until then.

app/models/krikri/provider.rb
no-reply: Ah, I see what you mean.  On the one hand, yes it seems worth it. On the other it doesn't look like the `QaReport` actually accesses the provider's label.

It seems like the extra Solr call only applies to providers built directly with `.new` rather than `.find` or `.all`.  I'd say make the change.

spec/models/provider_spec.rb
AudreyAltman: I am not sure I understand your comment.  If I change line 80 to:

```expect(described_class.all).to be_empty```

then the spec still passes.  Does that answer your question?  If so, I will commit that change.

Comment (no-reply) [inline spec/models/provider_spec.rb]: Yeah, that's the question. `.to be_empty` is slightly more specific, but mainly I wanted to make sure that this wasn't changed because it failed.

:+1: on changing it back.

Comment (AudreyAltman) [inline spec/models/provider_spec.rb]: Ahhh I see.  I didn't even notice I'd changed it  :)

**Comment by AudreyAltman** [comment]
I have addressed comments from @no-reply and :squashed:

File: spec/models/provider_spec.rb [hunk 7]
```diff
@@ -1,32 +1,41 @@
...
 
-  let(:provider) do
-    provider = Krikri::Provider.new(local_name)
-    provider.providedLabel = "moomin"
-    provider
-  end
+  let(:provider_base) { Krikri::Settings.prov.provider_base }
+  let(:id) { '123' }
+  let(:rdf_subject) { provider_base + id }
+  let(:name) { 'Snork Maiden Archives' }
```
no-reply: :moomin:

no-reply:
I'm :+1. Any objections to :merge: ?

markbreedlove:
no objection.

PR #178, opened 2015-07-13
Add enrichment to create prefLabels from provided

no-reply:
For most providers, we want to copy providedLabels into prefLabel prior
to running enrichments which will alter the label. This keeps the
providedLabels as a true representation of the original data.

Fields are ignored and returned as-is if:
   - they are not `ActiveTriples::Resource`
   - they do not respond to `#providedLabel`
   - they already have a `prefLabel`
   - there are no 'providedLabel`s present

### DIFF
--- a/lib/krikri/enrichments/create_pref_label_from_provided.rb
+++ b/lib/krikri/enrichments/create_pref_label_from_provided.rb
@@ -1,0 +1,58 @@
+module Krikri::Enrichments
+  ##
+  # Given an field that accepts both `providedLabel` and `prefLabel`, copies 
+  # the first `providedLabel` into `prefLabel` unless one is already present.
+  #
+  # Only the first `providedLabel` is copied, avoiding conflicts with SKOS's 
+  # limit of one `skos:prefLabel` per language tag.
+  #
+  # Fields are ignored and returned as-is if:
+  #   - they are not `ActiveTriples::Resource`
+  #   - they do not respond to `#providedLabel`
+  #   - they already have a `prefLabel`
+  #   - there are no `providedLabel`s present
+  # 
+  # @example enriching a resource with a providedLabel
+  #   label_enricher = CreatePrefLabelFromProvided.new
+  #
+  #   resource.providedLabel = 'moomin'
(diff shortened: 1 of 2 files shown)


anarchivist:
:+1: from me.

lib/krikri/enrichments/create_pref_label_from_provided.rb
```diff
@@ -0,0 +1,58 @@
...
+  #
+  # Only the first `providedLabel` is copied, avoiding conflicts with SKOS's 
+  # limit of one `skos:prefLabel` per language tag.
+  #
+  # Fields are ignored and returned as-is if:
+  #   - they are not `ActiveTriples::Resource`
+  #   - they do not respond to `#providedLabel`
+  #   - they already have a `prefLabel`
+  #   - there are no 'providedLabel`s present
+  #
```
AudreyAltman: This is so nit-picky, but you have a ' where I think you mean to have a ` in the line above.

no-reply: :+1: will fix.

lib/krikri/enrichments/create_pref_label_from_provided.rb
```diff
@@ -0,0 +1,58 @@
...
+  #   - they are not `ActiveTriples::Resource`
+  #   - they do not respond to `#providedLabel`
+  #   - they already have a `prefLabel`
+  #   - there are no 'providedLabel`s present
+  # 
+  # @example enriching a resource with a providedLabel
+  #   label_enricher = CreatePrefLabelFromProvided.new
+  #
+  #   resource.providedLabel = 'moomin'
+  #   enrich_value(resource)
```
AudreyAltman: Would this be `label_enricher.enrich_value(resource)`?

no-reply: yes

AudreyAltman:
Aside from my comments regarding comments (ha) this looks good to me.

no-reply:
@AudreyAltman's meta-comments have been addressed, and this is squashed for merge.

PR #179, opened 2015-07-16
Add `Activity#to_s` for easier activity display

no-reply:
Punts `#to_s` to `#inspect` to simplify activity discovery, e.g.:

      puts Krikri::Activities.all
      # the output is now equivalent to:
      # `puts Krikri::Activities.all.map(&:inspect)`

This was bothering me, now it shouldn't anymore.

### DIFF
--- a/app/models/krikri/activity.rb
+++ b/app/models/krikri/activity.rb
@@ -85,6 +85,8 @@ module Krikri
       RDF::URI(Krikri::Settings['marmotta']['ldp']) /
         Krikri::Settings['prov']['activity'] / id.to_s
     end
+    alias_method :to_term, :rdf_subject
+    
 
     ##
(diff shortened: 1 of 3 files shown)


markbreedlove:
This looks great.

Could you comment on the significance of `#to_term`?  I haven't noticed that before.

no-reply:
I obviously need to fix this up, per :travis: but while I do that...

`#to_term` is part of the standard interface for `RDF::Value`, which allows objects to be treated as values in an `RDF::Statement`. See: http://www.rubydoc.info/github/ruby-rdf/rdf/RDF/Value#to_term-instance_method

no-reply:
I rebased this.

anarchivist:
:+1:

PR #180, opened 2015-07-23
check for existence of provider_name

AudreyAltman:
This fixes an error that was causing the QA interface to crash if it tried to call `Krikri::Provider.name` on an indexed provider that was missing a value for `provider_name`.

### DIFF
--- a/app/models/krikri/provider.rb
+++ b/app/models/krikri/provider.rb
@@ -154,10 +154,11 @@ module Krikri
       query_params = { :rows => 1,
                        :q => rdf_subject,
                        :qf => 'provider_id',
-                       :fl => 'provider_name' }
+                       :fl => 'provider_name provider_id' }
       response = Krikri::SolrResponseBuilder.new(query_params).response
       return nil unless response.docs.count > 0
-      response.docs.first['provider_name'].first
+      response.docs.first['provider_name'].respond_to?(:first) ?
+        response.docs.first['provider_name'].first : rdf_subject
     end
 
     ##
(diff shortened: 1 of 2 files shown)


app/models/krikri/provider.rb
```diff
@@ -157,7 +157,8 @@ def initialize_name
                        :fl => 'provider_name' }
       response = Krikri::SolrResponseBuilder.new(query_params).response
       return nil unless response.docs.count > 0
-      response.docs.first['provider_name'].first
+      response.docs.first['provider_name'].respond_to?(:first) ?
+        response.docs.first['provider_name'].first : nil
```
anarchivist: Can we have this return the `id` instead of `nil`?

anarchivist:
Thanks, @AudreyAltman - minor comment, expanding on above - what will get returned and rendered in the nav if a `Krikri::Provider`'s name is `nil`?

app/models/krikri/provider.rb
AudreyAltman: Good idea!  This way, the `id` will get displayed in the nav if the provider's name is `nil`.

AudreyAltman:
I addressed the comment from @anarchivist and :squash:ed

app/models/krikri/provider.rb
no-reply: :+1: :clap: This is great. It solves the problem--for this case--of whether `nil` really means "nothing" or "this other thing that is slightly an odd issue" in a nicely encapsulated way.

no-reply:
:clap:

PR #181, opened 2015-07-24
field value report with csv download

AudreyAltman:
This introduces field value reports based on data that has been indexed into Solr.  The field value reports show all values for a given field in documents with a given provider.  It addresses the following tickets:
https://www.pivotaltracker.com/n/projects/1172184/stories/79608350
https://www.pivotaltracker.com/n/projects/1172184/stories/79330180
https://www.pivotaltracker.com/n/projects/1172184/stories/85081820

__CSV downloads__
This only offers the reports as CSV downloads - there is no web interface for the reports. Each row of the CSV table has three columns: the reporting field, the document `id`, and the document's `isShownAt_id` link.

If a document has a missing value for the reporting field, it will show up in the CSV table with `__MISSING__` as the field value.

If there are no indexed documents for the given provider, a CSV table will be generated with only one row, containing the headers.

__ActiveModel__
In this iteration of the code, `ActiveModelBase` is a superclass.  `FieldValueReport` and `Provider` have both been made subclasses of `ActiveModelBase`.  We may decide to change `ActiveModel` to a module, depending on the outcome of our conversation here: https://gist.github.com/AudreyAltman/002b75e291fc5c7f4ffe  This change can be made in a separate PR so that @guegueng can have access to these reports ASAP.

__Persistence__
In this iteration, field value reports are not persisted, but we may decide to persist them in the future if there are performance issues.

__Relationship between `FieldValueReport` and `Provider`__
The field value report route is nested under the provider route.  This model could be extended to all other reports and to records so we can do away with the `current_provider` strategy.

`FieldValueReports` and `Provider` have a relationship that approximates a `belongs_to` - `has_many` relationship, considering that `ActiveModel` does not support these relationships in the way that `ActiveRecord` does.  `FieldValueReport` has a `provider` attribute, and `Provider` has a (lazy-loading) `field_value_reports` attribute. 

I considered using the `activemodel-associations` gem to manage the relationship between `FieldValueReport` and `Provider`, but it turns out that this gem only works if you are associating an `ActiveModel` object with an `ActiveRecord` object, so it is not appropriate here.

__QaReport code__
I did not delete any of the QaReport code that @no-reply wrote b/c I wasn't sure if we would want to return to it any any time.  This code generates QA reports (including field value reports) from data that has been stored in Marmotta.

### DIFF
--- a/app/controllers/krikri/field_value_reports_controller.rb
+++ b/app/controllers/krikri/field_value_reports_controller.rb
@@ -1,0 +1,21 @@
+module Krikri
+  ##
+  # Handles HTTP requests for Field Value Reports
+  #
+  # @see Krikri::FieldValueReport
+  class FieldValueReportsController < ApplicationController
+    ##
+    # Renders the show view for the field value report, given by a compound key.
+    # The compound key is comprised by the field value report's field
+    # (represented by the route's id param) and the provider's id.
+
+    def show
+      @field_value_report = Krikri::FieldValueReport.find(params[:id], 
+                                                          params[:provider_id])
+
+      respond_to do |format|
+        format.csv
+      end
+    end
+  end
+end
--- a/app/models/krikri/field_value_report.rb
+++ b/app/models/krikri/field_value_report.rb
@@ -1,0 +1,124 @@
+module Krikri
+  ##
+  # FieldValueReport gives all unique values for a given field within a
+  # document.  It represents data that has been indexed into Solr.
+  #
+  # This is a read-only object.  FieldValueReports are not persisted.
+  class FieldValueReport < Krikri::ActiveModelBase
+    ##
+    # @!attribute :field [String] the name of the field.
+    #   @example: 'sourceResource_title'
+    #
+    # @!attribute :provider [Krikri::Provider]
+    attr_accessor :field, :provider
+
+    ##
+    # @param field [String] the name of the field being reported on
+    #   @example: 'sourceResource_title'
+    # @param provider_id [String] the id of the provider being reported on
+    # These two params act as a compound key.
+    #
+    # @return [Krikri::FieldValueReport]
+    def self.find(field, provider_id)
+      return nil unless fields.include? field.to_sym
+      provider = Krikri::Provider.find(provider_id)
+      return nil unless provider.present?
+      new({ :field => field,
+            :provider => provider })
+    end
+
+    ##
+    # All of the fields for which a report can be created.
+    def self.fields
+      [:dataProvider_providedLabel,
+       :sourceResource_alternative_providedLabel,
+       :sourceResource_collection_title,
+       :sourceResource_contributor_providedLabel,
+       :sourceResource_creator_providedLabel,
+       :sourceResource_date_providedLabel,
+       :sourceResource_description,
+       :sourceResource_format,
+       :sourceResource_genre_providedLabel,
+       :sourceResource_language_providedLabel,
+       :sourceResource_publisher_providedLabel,
+       :sourceResource_rights,
+       :sourceResource_rightsHolder_providedLabel,
+       :sourceResource_spatial_providedLabel,
+       :sourceResource_subject_providedLabel,
+       :sourceResource_temporal,
+       :sourceResource_title,
+       :sourceResource_type_name]
+    end
+
+    ##
+    # The headers for the report table, and the values to be returned from a
+    # Solr query.
+    def headers
+      [:id, :isShownAt_id, field.to_sym]
+    end
+
+    ##
+    # @param opts [Hash] optional parameters for the solr request
+    #   @example: enumerate_rows(batch_size: 1000)
+    # 
+    # @return Enumerator[<Array>] an enumerator over the rows
+    def enumerate_rows(opts = {})
+      Enumerator.new do |yielder|
+        loop do
+          opts = query_opts(opts)
+          response = Krikri::SolrResponseBuilder.new(opts).response
+          break if response.docs.empty?
+
+          parse_solr_response(response).each do |row|
+            yielder <<  headers.map { |header| row[header] }
+          end
+
+          opts[:start] += opts[:batch_size]
+          break if opts[:start] >= response.total
+        end
+      end
+    end
+
+    private
+
+    ##
+    # @param opts [Hash] additional options for the query params
+    # @return [Hash] parameters for a Solr request
+    def query_opts(opts = {})
+      { :fq => "provider_id:\"#{provider.rdf_subject}\"",
+        :fl => headers,
+        :rows => opts.fetch(:batch_size, 1000).to_i,
+        :start => opts.fetch(:start, 0).to_i }
+    end
+
+    ##
+    # @param [Blacklight::SolrResponse] response
+    # @return [Array<Hash>]
+    #
+    # The return Hashes should include keys for all headers defined in the
+    # :headers method
+    def parse_solr_response(response)
+      rows = []
+
+      response.docs.each do |doc|
+        id = doc['id'].split('/').last
+        isShownAt_id = doc['isShownAt_id'].respond_to?(:first) ?
+          doc['isShownAt_id'].first : '__MISSING__'
+
+        if doc[field].present?
+          doc[field].each do |value|
+            rows << { :id => id,
+                      field.to_sym => value,
+                      :isShownAt_id => isShownAt_id }
+          end
+        else
+          rows << { :id => id,
+                    field.to_sym => '__MISSING__',
+                    :isShownAt_id => isShownAt_id }
+        end
+      end
+
+      rows
+    end
+  end
+end
--- a/app/models/krikri/provider.rb
+++ b/app/models/krikri/provider.rb
@@ -6,18 +6,17 @@ module Krikri
   # This is a read-only object. It does not write new providers to Solr, nor
   # does it update or delete providers.
   #
+  # If Provider is initialized with a valid value for :rdf_subject, the model
+  # can extrapolate all other attributes.
+  #
   # A DPLA::MAP::Agent object can be constructed from an instance of this
   # ActiveModel object using the :agent method. The DPLA::MAP::Agent object
   # interacts with Marmotta, rather than Solr.
   #
-  #   @example Krikri::Provider.find(rdf_subject: 'http://blah.com/123').agent
+  #   @example Krikri::Provider.new(rdf_subject: 'http://blah.com/123').agent
   #     => [DPLA::MAP::Agent]
   #
-  class Provider
-    include ActiveModel::Validations
-    include ActiveModel::Conversion
-    extend ActiveModel::Naming
-
+  class Provider < Krikri::ActiveModelBase
     ##
     # @!attribute :rdf_subject [String] the URI identifying the provider
     #   @example: "http://blah.com/contributor/123"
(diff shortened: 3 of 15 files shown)


app/models/krikri/field_value_report.rb
```diff
@@ -0,0 +1,167 @@
...
+      table
+    end
+
+    ##
+    # @return Array[<Hash>]
+    def self.compile_data(name, provider)
+      data = []
+      query_params = { :fq => "provider_id:\"#{provider.rdf_subject}\"",
+                       :fl => ['id', 'isShownAt_id', name],
+                       :rows => 1000,
```
AudreyAltman: 1000 rows at a time may not be optimal - we will have to experiment with larger data sets.  If anyone has a good estimation of what the desired number of rows at a time would be, please let me know.  1000 is a blind guess.

app/models/krikri/provider.rb
```diff
@@ -69,7 +47,8 @@ def self.all
 
       response.facet_pivot['provider_id,provider_name'].map do |facet|
         rdf_subject = facet['value']
-        name = facet['pivot'].present? ? facet['pivot'].first['value'] : nil
+        name = facet['pivot'].present? ? facet['pivot'].first['value'] :
+          rdf_subject
```
AudreyAltman: This corrects an oversight in PR #180 and ensures that in all cases in which the `name` attribute is initialized, it will return the provider's full URI if the `name` value is nil.

app/models/krikri/provider.rb
```diff
@@ -89,8 +68,10 @@ def self.find(id)
                        :fl => 'provider_id provider_name' }
       response = Krikri::SolrResponseBuilder.new(query_params).response
       return nil unless response.docs.count > 0
+      name = response.docs.first['provider_name'].respond_to?(:first) ?
+        response.docs.first['provider_name'].first : rdf_subject
       new({ :rdf_subject => rdf_subject,
-            :name => response.docs.first['provider_name'].first })
+            :name => name })
```
AudreyAltman: See comment above https://github.com/dpla/KriKri/pull/181/files#r35464416

app/models/krikri/field_value_report.rb
anarchivist: Could we make this configurable with a parameter?

app/models/krikri/field_value_report.rb
```diff
@@ -0,0 +1,157 @@
...
+    # @!attribute :name [String] the name of the field.
+    #   @example: 'sourceResource_title'
+    #
+    # @!attribute :data [Array<Hash>} rows of values for the report.  Each row
+    # represents data about a document.
+    #   @example: [{ :id=>"a1b2c3",
+    #                :sourceResource_title=>"Moomin Story",
+    #                :isShownAt_id=> "http://example.org/items/4321" }]
+    #
+    # @!attribue :provider [Krikri::Provider]
```
anarchivist: minor typo

app/models/krikri/field_value_report.rb
```diff
@@ -0,0 +1,157 @@
...
+  #
+  # If FieldValueReport is initialized with valid values for :name and
+  # :provider, it can extrapolate :data.
+  #
+  class FieldValueReport < Krikri::ActiveModelBase
+    ##
+    # @!attribute :name [String] the name of the field.
+    #   @example: 'sourceResource_title'
+    #
+    # @!attribute :data [Array<Hash>} rows of values for the report.  Each row
```
anarchivist: minor typo

anarchivist:
Looks good to me, at least in terms of a Friday afternoon review; I'm hoping @no-reply can have a look on Monday. A couple of questions for thought which don't necessarily need to be addressed now:

1) What are the performance implications for doing the reports this way?
2) How long will it take to generate a report of 80K records? 300K?, etc.? 
3) Should the report generation be handed off to a job that gets placed on a queue?

app/models/krikri/field_value_report.rb
AudreyAltman: :+1:

AudreyAltman:
@anarchivist re: your questions - I don't know, but if you have any suggestions for improving performance, queuing, etc. I would be interested to hear.

AudreyAltman:
I addressed the suggestions from @anarchivist.

app/controllers/krikri/field_value_reports_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+    ##
+    # Renders the show view for the field value report, given by a compound key.
+    # The compound key is comprised by the field value report's name
+    # (represented by the route's id param) and the provider's id.
+    def show
+      @field_value_report =
+        Krikri::FieldValueReport.find(params[:id], params[:provider_id])
+
+      respond_to do |format|
+        format.html
```
markbreedlove: Is this responding with a successful status and an empty response body?  (I'm looking at what I understand to be the corresponding view for HTML.)

Should we get an error instead if we're requesting HTML when only CSV makes sense?

AudreyAltman: :+1:

markbreedlove:
@AudreyAltman: sorry if any of this is rambling or unclear. Let me know if I can clarify any of it!

I found myself wondering, when I first approached this, whether the
`FieldValueReport` model was for a report itself (a document with a title and
and numbers, for example) or for the thing that's reported on.  It's actually used to
provide an array of things that are reported on, not the report itself,
so I think it could be renamed to `FieldValue`.  The `.find` method is passed
a `provider_id` parameter, which is used to find the correct records, but
which is not used subsequently, though `.new` is given it as an option.
I don't know if that's an ActiveModel requirement, but I found myself looking
at the call to `.new` in `.find`, trying to figure out what `provider` was
going to be used for.

I would usually expect the controller to produce a result handle for some field-value
model and hand that off to a view, which would be responsible for lazily iterating over
that and printing the result rows.

I see the responsibility of looping through the records and writing lines to be
that of the view.  Output should ideally be unbuffered, something that I assume
we're able to control in the controller.  This would reduce the memory
requirement of this operation.  The FieldValueReport model currently loads a
large array into memory before it gets written out.

`FieldValueReport#csv` instantiates `CSV::Row` for every row, when `Array`'s builtin
`#to_csv` seems to suit our needs.  For example,
```
[6] pry(main)> puts ['"quoted"', 'boring string with no punctuation', 'with,commas'].to_csv
"""quoted""",boring string with no punctuation,"with,commas"
```

Using `CSV::Table` to output a header line, plus the record lines,
takes more effort to follow and is probably less efficient than using
`#to_csv` calls on a simple array of strings.

This boils down to:  I think that unbuffered i/o with a view that iterates lazily over filtered `Krikri::SolrResponseBuilder` responses would be the most efficient thing.  If this gets worked into a job that is enqueued, we will have to build more endpoints and do some more u.i. Javascript work to handle asychronous operations.  This outlines the basic way it should work: http://restcookbook.com/Resources/asynchroneous-operations/  The class that writes the reports should still avoid loading the result set into memory as an array, and should lazily write each row as it's retrieved from Solr.

AudreyAltman:
Thanks for this excellent feedback, @markbreedlove !

AudreyAltman:
@markbreedlove I have made some progress on this PR.  It uses `#to_csv` instead of `CSV::Table`, and the view iterates lazily over an enumerator.  I haven't had a chance to renew my search for information about unbuffered i/o.  

I haven't changed the name of the `FieldValueReport` model.  I'm hoping that the name makes more sense with these edits.  The `.find` method may be awkward, but from what I've read, it seems like a common convention for a `ActiveModel`.  What I like about the `.find` method (as opposed to simply using a `new` method to instantiate the object) is that `.find` checks to make sure that the parameters are valid before instantiating the object - if the provider hasn't been indexed, or the name of the field being reported on isn't valid, it just returns nil.  This simplifies things b/c otherwise you would have to include checks in all the instance methods to make sure the provider and field name were present and valid before trying to use them.

Let me know if you have any questions.  We may need to iterate on this again, but I thought it would be worth getting your feedback first.

Final note: I don't understand why the Travis build is failing.

app/views/krikri/field_value_reports/show.csv.erb
```diff
@@ -0,0 +1 @@
+<%= @field_value_report.headers.to_csv %><% @field_value_report.enumerate_rows.each do |row| %><%= row.to_csv %><% end %>
```
AudreyAltman: This has no line breaks b/c they show up in the CSV file as blank lines.  Sorry, I know it's not nice to read.

anarchivist:
> Final note: I don't understand why the Travis build is failing.

I think the issue is the same one that #184 is set to address - checking out `develop`, running `git pull`, and switching back to your feature branch and rebasing should address the build failure.

AudreyAltman:
Tests are now passing!  Thanks @anarchivist :D

anarchivist:
Looks like there was a minor issue with the rebase - I'm going force push a re-rebased set of changes...

anarchivist:
Alright, I'm feeling good about this. @markbreedlove, @no-reply - y'all want to look on Wednesday?

app/models/krikri/field_value_report.rb
```diff
@@ -0,0 +1,124 @@
...
+          opts = query_opts(opts)
+          response = Krikri::SolrResponseBuilder.new(opts).response
+          break if response.docs.empty?
+
+          parse_solr_response(response).each do |row|
+            yielder <<  headers.map { |header| row[header] }
+          end
+
+          opts[:start] += opts[:rows]
+          break if opts[:start] >= response.total
```
markbreedlove: Does this work if `response.total` is not divisible by `opts[:rows]`?

app/controllers/krikri/field_value_reports_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+    # Renders the show view for the field value report, given by a compound key.
+    # The compound key is comprised by the field value report's field
+    # (represented by the route's id param) and the provider's id.
+
+    def show
+      @field_value_report = Krikri::FieldValueReport.find(params[:id], 
+                                                          params[:provider_id])
+
+      respond_to do |format|
+        format.csv
```
markbreedlove: This appears to be designed to coerce the Content-Type to text/csv.  Is that correct?

app/models/krikri/field_value_report.rb
markbreedlove: Never mind.  I walked through it, and it looks like it does the right thing.

app/controllers/krikri/field_value_reports_controller.rb
markbreedlove: That was just a curious question.  I've verified since then that this is how you make sure that the output is CSV, regardless of the type requested.

markbreedlove:
This looks fine.

There is an option `:rows` that is passed to `Krikri::FieldValueReport#enumerate_rows` that represents a batch size.  Elsewhere in the model, there is a variable named `rows` that is an array of rows.  This gave me a momentary pause, and I think that the batch size could be renamed to `:batch_size`.

We may want to play around with the default batch size in the future, and make it overridable by a configuration variable, but I think we should try to run with this now and see how it works out.

markbreedlove:
I took the liberty of changing that `:rows` parameter, because @AudreyAltman is on vacation.

markbreedlove:
Just a minor note about the unbuffered I/O, because I forgot to follow up on it.  We're going to see how it works as it is.  If we think memory usage or performance needs to be improved, we can spend time on it later.

PR #182, opened 2015-07-27
Display only the `@graph` in JSON-LD displays

no-reply:
Aggregations previously displayed with full context. This displays them with only the actual records.


no-reply:
I force pushed this with a minor fix that caused the previous failure.

anarchivist:
:rocket:

PR #183, opened 2015-07-30
Add prefLabels to languages

no-reply:
We prefer the last rdf:Label from [RDF::ISO_639_3](https://raw.githubusercontent.com/dpla/dpla_map/map-4.0/lib/rdf/iso_639_3.rb). I.e. "French" rather than "French language" and "Creole, French" rather than "French Creole".


AudreyAltman:
Looks good to me.

PR #184, opened 2015-08-06
Pin rails_config to version 0.4.0

markbreedlove:
Pin rails_config to version 0.4.0 to prevent the error
"NameError: uninitialized constant Settings."

https://rubygems.org/gems/rails_config/versions/0.99.0 says "Please
install the Config gem instead."


no-reply:
:+1:

PR #185, opened 2015-08-11
Corrects ending punctuation enrichment

no-reply:
The ending punctuation enrichment was missing long words that ended with '.' due to an error in the regexp that was intended to keep it from removing the trailing '.' on initials. This fixes that by changing `*` to `+`.


anarchivist:
:rocket:

PR #186, opened 2015-08-12
Update DPLA::MAP dependency to -pre.11

no-reply:
This updates DPLA::MAP to deal with the language handling issue we saw with previous enrichhments.


anarchivist:
:+1:

PR #187, opened 2015-08-12
Fix whitespace bug in RemoveEmptyFields

no-reply:
A Regex error caused the enrichment to remove fields beginning with whitespace + a newline, even if there was other content present.


anarchivist:
awesome

PR #188, opened 2015-08-14
Fix FieldValueReport error where nil can't be coerced into Fixnum (fixes #7883)

anarchivist:
This was introduced by trying to use the admittedly clearer key `:batch_size`.


no-reply:
:rocket:

PR #190, opened 2015-08-18
Upgrade DPLA::MAP

no-reply:
This upgrades to DPLA::MAP pre.12. It removes the dependency on LinkedVocabs, which is deserving of a rewrite against the latest `ActiveTriples` work and updated BNode handling.


AudreyAltman:
This looks good to me, but @markbreedlove should probably take a took, too.

PR #191, opened 2015-08-25
Handle errors in indexer

no-reply:
Instead of catching everything at the top level Indexer, we now handle
errors within the index implementation. Anything that reaches the
Indexer is treated as an failure state and logged accordingly, then
reraised.

### DIFF
--- a/app/models/krikri/activity.rb
+++ b/app/models/krikri/activity.rb
@@ -18,7 +18,6 @@ module Krikri
     #                   the activity
     # @!attribute start_time
     #    @return [DateTime] a datestamp marking the activity's start
-
     validate :agent_must_be_a_software_agent
 
     def agent_must_be_a_software_agent
@@ -40,18 +39,26 @@ module Krikri
     ##
     # Runs the block, setting the start and end time of the run. The given block
     # is passed an instance of the agent, and a URI representing this Activity.
+    # 
+    # Handles logging of activity start/stop and failure states.
+    #
+    # @raise [RuntimeError] re-raises logged errors on Activity failure
     def run
       if block_given?
         update_attribute(:end_time, nil) if ended?
+        Krikri::Logger
+          .log(:info, "Activity #{agent.constantize}-#{id} is running")
         set_start_time
         begin
           yield agent_instance, rdf_subject
         rescue => e
-          Rails.logger.error("Error performing Activity: #{id}\n" \
-                             "#{e.message}\n#{e.backtrace}")
+          Krikri::Logger.log(:error, "Error performing Activity: #{id}\n" \
+                                     "#{e.message}\n#{e.backtrace}")
           raise e
         ensure
           set_end_time
+          Krikri::Logger
+            .log(:info, "Activity #{agent.constantize}-#{id} is done")
         end
       end
     end
--- a/lib/krikri/logger.rb
+++ b/lib/krikri/logger.rb
@@ -1,0 +1,38 @@
+module Krikri
+  ##
+  # An application-wide tagged logger
+  # 
+  # @example with default logger
+  #   Krikri::Logger.log :info, 'message'
+  #
+  # @example with custom logger
+  #   logger = Logger.new(my_logger)
+  #   logger.log :info, 'message'
+  class Logger
+    ##
+    # @param [ActiveSupport::TaggedLogging] logger
+    def initialize(logger = ActiveSupport::TaggedLogging.new(Rails.logger))
+      @logger = logger
+    end
+
+    class << self
+      ##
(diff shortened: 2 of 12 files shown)


no-reply:
Looking a bit closer, it seems like it might be reasonable to the error handling out of `Indexer#run` altogether. `Activity#run` already catches, logs, and reraises errors.

no-reply:
Likewise, we may consider moving logging. Proposed changes incoming as a separate commit.

lib/krikri/search_index.rb
```diff
@@ -116,6 +117,17 @@ def entities_as_json_hashes(activity)
...
       aggregation.to_jsonld['@graph'][0]
     end
+
+    private
+    
+    def index_with_error_handling(activity, &block)
+      begin
+        yield if block_given?
+      rescue => e
+        log :error, "indexer error for #{activity}:"
```
anarchivist: Minor nit: consider `log :error, "indexer error for Activity #{activity}:"` instead

lib/krikri.rb
```diff
@@ -7,7 +7,7 @@
 require 'audumbla'
 require 'krikri/engine'
 
-module Krikri
+module Krikri
```
anarchivist: Minor style nit: remove trailing whitespace

lib/krikri/logger.rb
```diff
@@ -0,0 +1,38 @@
...
+      # @see #log
+      def log(priority, msg)
+        new.log(priority, msg)
+      end
+    end
+
+    ##
+    # Log a message, tagged for application-wide consistency
+    #
+    # @param [Symbol] priority  a priority tag
```
anarchivist: Minor nit: `priority` is not a tag; probably better to just use `a priority level`.

lib/krikri/logger.rb
```diff
@@ -0,0 +1,38 @@
...
+  # @example with default logger
+  #   Krikri::Logger.log :info, 'message'
+  #
+  # @example with custom logger
+  #   logger = Logger.new(my_logger)
+  #   logger.log :info, 'message'
+  class Logger
+    ##
+    # @param [ActiveSupport::TaggedLogging] logger
+    def initialize(logger = ActiveSupport::TaggedLogging.new(Rails.logger))
```
anarchivist: I understand the intent here (i.e., to be able to potentially swap in other tagged logging implementations), but is the assumption that anything would expose a similar interface as `ActiveSupport::TaggedLogging`, based on what's on [line 34](https://github.com/dpla/KriKri/pull/191/files#diff-31e82b2b381da2aeda2c21013a337c1dR34) below?

no-reply: Right. The yard doc for the parameter also asserts this. You could do, e.g. `Logger.new(ActiveSupport::TaggedLogging.new(some_logger)`, or implement a null logger that just eats the messages, or whatever. :)

anarchivist: I somehow managed to overlook the YARD annotations here when I first reviewed it.

app/models/krikri/activity.rb
```diff
@@ -40,18 +39,24 @@ def set_end_time
...
     # Runs the block, setting the start and end time of the run. The given block
     # is passed an instance of the agent, and a URI representing this Activity.
+    # 
+    # Handles logging of activity start/stop and failure states.
+    #
+    # @raise [RuntimeError] re-raises logged errors on Activity failure
     def run
       if block_given?
         update_attribute(:end_time, nil) if ended?
+        Krikri::Logger.log :info, "Activity #{id} is running"
```
markbreedlove: Could this also log the value of `agent.contantize` so that it's easy to tell in the logfile what kind of activity this is?

lib/krikri/search_index.rb
no-reply: This is also missing `Krikri::Logger`, adding a test and fixing the text as requested.

markbreedlove:
I think this looks great.  I don't have any further observations.

lib/krikri/search_index.rb
```diff
@@ -33,11 +33,10 @@ def bulk_add(_)
     # Shim that determines, for a particular type of index, which strategy
     # to use, adding a single document, or adding them in bulk.  Intended
     # to be overridden as necessary.
+    #
     # @see #add
     # @see #bulk_add
     def update_from_activity(activity)
-      fail "#{activity} is not an Activity" \
```
no-reply: Note that I've simply removed this, relying on duck typing instead.  Anything with `#entities` will work fine, here.

no-reply:
This is rebased.

PR #192, opened 2015-08-25
Fix prefLabel issue in map_crosswalk

no-reply:
Fixes a typo on `prefLabel` which caused failure on index for certain records.

Adds a regression test.


anarchivist:
Looks good. I'm guessing this addresses the <code>Indexing error: undefined method `each' for nil:NilClass</code> error we're seeing.

anarchivist:
Never mind that second comment - I think that's in the QA indexer.

PR #194, opened 2015-08-27
Fix map crosswalk language

no-reply:
Maps language to a structure like:

  { name: 'English', iso639_3: 'eng' }

Also fixes a minor bug on subject terms, allowing us to index prefLabels
with preference over providedLabels.

### DIFF
--- a/lib/krikri/map_crosswalk.rb
+++ b/lib/krikri/map_crosswalk.rb
@@ -120,7 +120,7 @@ module Krikri
         set_value(sr, :identifier, parent_sr.identifier)
 
         set_value(sr, :language, parent_sr.language) do |lang|
-          get_label(lang)
+          build_language(lang)
         end
 
         set_value(sr, :publisher, parent_sr.publisher) do |pub|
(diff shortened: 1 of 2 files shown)


lib/krikri/map_crosswalk.rb
```diff
@@ -212,7 +228,7 @@ def build_subject(source)
         subject = {}
         subject[:name] = source.prefLabel.first if
           source.prefLabel.any?
-        subject[:name] = source.providedLabel.first if
+        subject[:name] ||= source.providedLabel.first if
```
AudreyAltman: :+1:

AudreyAltman:
So just to confirm:  If the source _has_ and name but _does not have_ and exact match, you want `:build_language` to return something like { name: 'English' }, correct?

no-reply:
Bingo. This requires a `name`, but only includes an `iso639_3` if there is an exact match. Anything with an exact match should also have a name, but I think it's a good idea to drop any edge cases, rather than having label-less languages.

AudreyAltman:
In that case, I think this is ready to :squash: and :merge:

no-reply:
Single commit, here; :merge: away!

AudreyAltman:
Oh, I see there is no :squash:ing necessary.  Shall I :merge:, or you do want someone else to take a look first?

PR #195, opened 2015-08-31
Add enrichment to clear WebResource Blank Nodes

no-reply:
`edm:WebResource` nodes should __always__ be an HTTP URI. This removes them if a blank node is given.


anarchivist:
:+1: Looks good to me.

PR #196, opened 2015-08-31
Merge index options with defaults

no-reply:
Passing in an empty options hash overwrote the default options, so jobs
would never use the defaults. This merges the options.

A side effect is that we avoid calling `#delete` on the original options
hash, leaving it intact for the caller.


anarchivist:
:rocket:

PR #197, opened 2015-09-04
Monkeypatch parser Scanner to catch long terminals

no-reply:
This commit should be reverted when https://github.com/gkellogg/ebnf/issues/5 is suitably closed.

### DIFF
--- a/lib/krikri.rb
+++ b/lib/krikri.rb
@@ -19,5 +19,27 @@ module Krikri
            'krikri/entity_behaviors/aggregation_entity_behavior'
   autoload :OriginalRecordEntityBehavior,
            'krikri/entity_behaviors/original_record_entity_behavior'
+end
 
+##
+# Monkey-patch the EBNF Scanner to catch larger terminals.
+#
+# @see https://github.com/gkellogg/ebnf/issues/5
+module EBNF::LL1
+  class Scanner
+    def initialize(input, options = {})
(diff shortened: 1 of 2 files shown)


anarchivist:
Do we need a test for this?

no-reply:
I was thinking not, but I'm open to argument.

The test would be something like the code seen in the issue linked above:

```ruby
str = "+#{'0' * 10000}\n+"
s = EBNF::LL1::Scanner.new(StringIO.new(str))
expect(s.scan(/\+.*\+/m)).not_to be_nil
```

My impulse is that we should include the test only if we would retain it as a regression test when the upstream code is fixed and this change is removed.  We're parsing turtle throughout the test suite, so we can be confident that this doesn't break normal cases.

anarchivist:
> My impulse is that we should include the test only if we would retain it as a regression test when the upstream code is fixed and this change is removed.

In my mind, I think this is the argument for adding such a regression test. Let me if you think otherwise!

no-reply:
Maybe what we need is a test closer to the surface?

I'm thinking that the test above may or may not pass, depending on implementation. But something that uses a real turtle example and runs through the turtle parser would be of greater value.

anarchivist:
Works for me.

no-reply:
There you go. I've confirmed that this fails without the patch in place. 

We should keep the commits separate, so the first can be reverted simply when a fix comes through on `ebnf`.

PR #198, opened 2015-09-10
Error handling in Mapping

no-reply:
Introduces a `Mapping::Error` that tracks multiple errors for a given
record and handles messaging.

In the `Mapping`, we catch errors for each declaration on a given
record, storing them with the relevant property name, and re-raise. The
`Mapper` then simply catches the error and logs its message.

This simplifies handling outside the DSL code, and creates richer error
messages.

### DIFF
--- a/lib/krikri/mapping.rb
+++ b/lib/krikri/mapping.rb
@@ -8,6 +8,24 @@ module Krikri
   #    map.process_record(my_original_record)
   #    # => #<MyModelClass:0x3ff8b7459210()>
   #
+  # When one or more errors are encoutered during processing, they are collected
+  # in a `Krikri::Kapping::Error` and re-raised.
+  #
+  # @example When an error is thrown during property mapping
+  #    map = Mapping.new(MyModelClass)
+  #    begin
+  #      map.process_record(my_original_record)
+  #    rescue Mapping::Error => e
+  #      e.message
+  #    end
+  #    # => Property failed on subject:
+  #    #       {subject error message}
+  #    #       {subject error backtrace}
+  #    #    Property failed on title:
+  #    #       {title error message}
+  #    #       {title error backtrace}
+  #   
+  # @see Krikri::MappingDSL
   class Mapping
     include MappingDSL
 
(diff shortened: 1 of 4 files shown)


AudreyAltman:
Just to check - since you're defining `Error` as in inner class of `Mapping`, does that mean that you don't anticipate using this type of error handling for any other classes?

AudreyAltman:
My question above is mostly academic.  This is really neat!  :+1: from me.

no-reply:
> does that mean that you don't anticipate using this type of error handling for any other classes?

Yeah. This is a pattern we could deploy elsewhere, but the specific behavior is all particular to mapping.

We might benefit from some digging to find other people using error collector patterns in Ruby---and think about a refactor to use a base class if we find ourselves reaching for this again. 

Since implementing this, I've already found references to this kind of code in the Fowler DSL book. Everything old is new again.

spec/lib/krikri/mapping_spec.rb
```diff
@@ -28,15 +28,52 @@
...
       subject { described_class.new(target_class, parser, *parser_args) }
+      let(:target_instance) { double('target instance') }
+    end
+    
+    describe 'error handling' do  
+      include_context 'property declarations'
+
+      before do
+        allow(target_instance).to receive(:error_property=).and_raise error
+        subject.error_property 'moomin'
```
markbreedlove: Does this get executed after `subject.process_record(record)`, below?
I'm just curious how this is is invoked.  Is this calling `:error_property=`?

markbreedlove:
I can generally see what's going on by reading the tests, but maybe there could be some `@example`s in the doc comments.

lib/krikri/mapping.rb
```diff
@@ -33,10 +33,56 @@ def initialize(klass = DPLA::MAP::Aggregation,
...
+      end
+      
+      ##
+      # @return [String] a message describing the full error set
+      def message
+        msg = "Error processing mapping for #{original_record.local_name}\n"
+        errors.each do |property, error|
+          msg << "Failed on property #{property}:\n"
+          msg << "\t#{error.message}\n\t#{error.backtrace.join("\n\t")}"
+        end
```
markbreedlove: The output of `#message` will be great to have.

spec/lib/krikri/mapping_spec.rb
no-reply: >  Is this calling `:error_property=`?
Exactly. `subject.error_property value` creates a `PropertyDeclaration` in the mapping. When the mapping is executed it calls `#error_property=` with `value` on the `target_instance`, in this case invoking the error we want to handle.

no-reply:
I'll work on adding some examples. :+1: for pushing on docs. :)

no-reply:
Does that look better?

markbreedlove:
Yes, thanks!

markbreedlove:
Happy to merge; assuming you want to squash it first.

no-reply:
:squash:'d

PR #199, opened 2015-10-23
Feature/ttl update

no-reply:
This version of the Turtle parser uses a hand-written parser that is about 3x faster than the previous version. 

It also addresses the issue with long literals. The monkey patch for that issue is reverted in c0c4448; the regression test added by 0f9a1b7 remains.


anarchivist:
:+1:

PR #200, opened 2015-11-03
Add if/else for `Parser::ValueArray`

no-reply:
For initial review:

```ruby
value_array.if { |arry| arry.field(:a_field) }
  .else { |arry|  arry.field(:alternate_field) }

value_array.if.field(:a_field).reject { |v| v == 'SKIP ME' }
  .else { |arry|  arry.field(:alternate_field) }
```

This addresses the use case in: https://github.com/dpla/heidrun-mappings/commit/5768026c078ab1a037d7eb043f4abd9b72a9f289#commitcomment-13144153

### DIFF
--- a/lib/krikri/parser.rb
+++ b/lib/krikri/parser.rb
@@ -141,20 +141,36 @@ module Krikri
     # A specialized Array object for containing Parser::Values. Provides methods
     # for accessing and filtering values that can be chained.
     #
+    # @example chaining methods to select values
+    #
     #     my_value_array.field('dc:creator', 'foaf:name')
     #       .match_attribute('first_name').values
     #
     # Methods defined on this class should return another ValueArray, an Array
     # of literal values (retrieved from Parser::Value#value), or a single
     # literal value.
+    #
+    # Uses `@top` to track a base node for recovery via `#else`. Methods that 
+    # return a `ValueArray` should pass `@top` down to the new instance.
+    #
+    # @example conditional recovery. `#if` sets `@top`, `#else` rewinds if empty
+    #
+    #     my_value_array.field('dc:creator', 'foaf:name').if.field('empty:field')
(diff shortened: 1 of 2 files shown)


no-reply:
The failures are due to dependency issues. Taking a look now.

lib/krikri/parser.rb
```diff
@@ -212,7 +215,55 @@ def fields(*args)
...
           field(*Array(f))
         end
-        self.class.new(results.flatten)
+        self.class.new(results.flatten, @top)
+      end
+
+      ##
+      # Sets the top of the call chain to self and returns or yields self
+      #
+      # @example with method chain syntax
```
anarchivist: :tada: great docs :memo:

anarchivist:
Looks good. I think this tests should pass if you rebase on `develop`.

anarchivist:
Meanwhile, assigning to @markbreedlove for a second look.

markbreedlove:
Looks good to me.  The only thing I can think of at the moment is that it might be good to document the significance of `@top` in `ValueArray#initialize` or maybe the class itself, because of its importance.

no-reply:
If the new docs look good, and we aren't concerned about handling for extra complex cases, this is ready.

Are we worried about testing, e.g.:

```ruby
some_node.field('a', 'b', 'c').else { |arry| arry.field('d', 'e') }.else { |arry| arry.field('f', 'g') }
some_node.if.field('a', 'b', 'c').if.reject { |v| is_bad(v) }.else { |arry| arry.field('d', 'e') }
```

And other such aberrations people may try?

markbreedlove:
I don't think extra tests for those would be necessary.  This should be fine to squash.

no-reply:
I'll add some further notes to the commit message, if you want to hold off a moment.

markbreedlove:
ok

no-reply:
done.

PR #202, opened 2015-11-04
Force encoding on RDFSource response bodies

no-reply:
`Net::HTTP` is returning `ASCII-8BIT` strings for some responses (see [this code](https://github.com/ruby/ruby/blob/trunk/lib/net/http/response.rb#L373-L377) and [this issue](https://bugs.ruby-lang.org/issues/2567)). This forces `RDFSource` responses (which for LDP MUST be `text/turtle` and therefore UTF-8, if requested) to read as UTF-8.

We considered using middleware like [`Faraday::Encoding`](https://github.com/ma2gedev/faraday-encoding) but Marmotta does not return an explict encoding in `content-type` headers.

### DIFF
--- a/lib/krikri/ldp/rdf_source.rb
+++ b/lib/krikri/ldp/rdf_source.rb
@@ -73,11 +73,18 @@ module Krikri::LDP
     private
 
     ##
-    # Clears the RDF::Graph and repopulates it from the http body.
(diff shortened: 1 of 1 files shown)


markbreedlove:
Looks good to me.

anarchivist:
Can we add a line of documentation? :) Otherwise :+1:

no-reply:
Docs added.

PR #203, opened 2015-11-09
Allow OAI to recover on errors when iterating sets

no-reply:
Adds error recovery when an attempt to retrieve a set throws an
`OAI::Excepton`. OAI-PMH requests throw errors when no matches can be
found with (e.g.) the required metadata prefix. When this happens we
want to pass over the set, logging & returning empty, instead of
aborting the harvest.

Note that the failover applies to *any* OAI exception. We've observed
that some provider endpoints report different errors, and
`OAI::Exception` does little to help us identify error types. This may
turn out to be heavy-handed.

### DIFF
--- a/spec/lib/krikri/harvesters/oai_harvester_spec.rb
+++ b/spec/lib/krikri/harvesters/oai_harvester_spec.rb
@@ -329,6 +329,21 @@ EOM
                                        .and_return(result)
             subject.send(method, { :skip_set => 'moomin' }).first
           end
+
+          it 'skips sets that error' do
+            args[:oai].delete(:set)
+            invalid = args[:oai].dup
+            invalid[:set] = 'invalid'
+            valid = args[:oai].dup
+            valid[:set] = 'valid'
+
+            allow(subject).to receive(:sets).and_return(['invalid', 'valid', 'moomin'])
+            expect(subject.client).to receive(request_type).with(invalid)
+                                       .and_raise(OAI::Exception, '')
+            expect(subject.client).to receive(request_type).with(valid)
+                                       .and_return(result)
+            subject.send(method, { :skip_set => 'moomin' }).first
+          end
         end
       end
 
--- a/lib/krikri/harvesters/oai_harvester.rb
+++ b/lib/krikri/harvesters/oai_harvester.rb
@@ -156,12 +156,18 @@ module Krikri::Harvesters
 
     ##
     # Runs the request in the given block against the sets specified in `opts`.
-    # Results are concatenated into a single enumerator
+    # Results are concatenated into a single enumerator.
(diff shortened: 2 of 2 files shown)


anarchivist:
:+1:

spec/lib/krikri/harvesters/oai_harvester_spec.rb
```diff
@@ -329,6 +329,21 @@
...
+            valid = args[:oai].dup
+            valid[:set] = 'valid'
+
+            allow(subject).to receive(:sets).and_return(['invalid', 'valid', 'moomin'])
+            expect(subject.client).to receive(request_type).with(invalid)
+                                       .and_raise(OAI::Exception, '')
+            expect(subject.client).to receive(request_type).with(valid)
+                                       .and_return(result)
+            subject.send(method, { :skip_set => 'moomin' }).first
+          end
```
markbreedlove: Do you think it's OK that there are two expectations in this spec?

Does it test that the result of `#request_with_sets` concatenates the results of successful sets?

markbreedlove:
This makes sense to me. Looks good.  Just had those minor questions about the test.

spec/lib/krikri/harvesters/oai_harvester_spec.rb
no-reply: > Do you think it's OK that there are two expectations in this spec?

Only sort of... it's not ideal.

I thought about refactoring to use a custom matcher, but the likely shape of that matcher didn't really pop out to me when I thought about it.

Alternatively, we could conform with the letter of the law (but maybe less the spirit) by changing the first `expect` to `allow`.  That would test the skipping more precisely, but would be a bit more fragile in that it would pass if for some reason the "invalid" set was just never called.

markbreedlove: I guess that I'm not very bothered by the two `expect`s.  If `concat_enum` (in `#request_with_sets`) is tested somewhere than it doesn't seem the place here to test that it works. I just haven't seen it anywhere else and was wondering if the concatenation is verified.  I see references for `collect_concat` / `flat_map`, but not this method. If this test asserts "the errored set is left out but the good one's element is included" then it should be OK.

no-reply:
@markbreedlove I partially answered your testing question.

I'm inclined to argue for pushing this through, though my mixed feelings are evident. :)

What seems best to you?

markbreedlove:
I think it's fine.

markbreedlove:
@AudreyAltman  is still assigned, so I'll just let her comment

AudreyAltman:
:+1:

no-reply:
:rocket:

PR #204, opened 2015-11-11
Introduce a MarcXMLHarvester in support of the Hathi harvester

payten:
`MarcXMLHarvester` defines the enumerators for parsing records from a MARC XML document.  Our new `HathiHarvester` uses this.


lib/krikri/harvesters/marc_xml_harvester.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    ##
+    # @return [Enumerator::Lazy] an enumerator of the records targeted by this
+    #   harvester.
+    def records
+      enumerate_records.lazy.map { |rec| build_record(rec) }
+    end
+
+    ##
+    # Gets a single record with the given identifier from the API
```
no-reply: This looks like a doc for `#get_record` rather than `#record_ids`.

lib/krikri/harvesters/marc_xml_harvester.rb
```diff
@@ -0,0 +1,106 @@
+module Krikri::Harvesters
+
+  class MarcXMLHarvester
```
no-reply: Can we add some class level docs for implementation requirements?  

Seems like a subclass will have to implement `#each_collection`, so maybe this should ship with an explicitly abstract method throwing `NotImplementedError`. That is the pattern we have followed for abstract methods to date.

lib/krikri/harvesters/marc_xml_harvester.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    def build_record(marcxml_doc)
+      @record_class.build(mint_id(marcxml_doc.identifier),
+                          marcxml_doc.source,
+                          content_type)
+    end
+  end
+
+
+  class MarcXMLDoc
```
no-reply: :+1: maybe some inline docs here, too?

lib/krikri/harvesters/marc_xml_harvester.rb
```diff
@@ -0,0 +1,106 @@
+module Krikri::Harvesters
+
+  class MarcXMLHarvester
+    include Krikri::Harvester
+
+    attr_reader :opts
+
+    def initialize(opts = {})
+      @max_records = opts.fetch(:max_records, :unlimited)
```
no-reply: Is there a use case for `:max_records` as an option? 

Typically when we want to harvest a small number of records we do `harvester.records.take(n)`, relying on the `#lazy` factor. Obviously that can't be passed in as an option to the job; if we think there's a good reason to do that, in general, perhaps we should add this to the base `Harvester` using something like the `#take(n)` implementation.

no-reply:
I added a branch at https://github.com/dpla/KriKri/tree/feature/tj-hathi with some added docs and specs. I'm working on verifying that the Heidrun PR passes those specs.

lib/krikri/harvesters/marc_xml_harvester.rb
markbreedlove: I'm in favor of this being an option, though probably in `Krikri::Harvester` as @no-reply suggests.

It would be convenient to be able to limit an enqueued harvest.

anarchivist: :+1: for #take the `Krikri::Harvester` level

no-reply: I'll implement that and send a PR, then maybe these can be rebased on the generalized solution.

marktriggs:
Thanks @no-reply.  We'd started on the documentation locally before we saw your branch, so we've merged your suggested changes in by hand (including the rspec files).

We've taken the `:max_records` option out for now, on the assumption that it'll get added to the top-level harvester instead.

markbreedlove:
Thanks for this! I don't think I have any further questions. I'll assign it to Mark M to pass it around.

marktriggs:
Thanks again @markbreedlove.  I've addressed those issues now.  I'll mark my other outstanding PRs as "Do not merge" while I review those other branches with an eye for the same issues.

anarchivist:
:+1: I think this is ready to be :squash:'ed.

marktriggs:
Squished! :elephant:

no-reply:
:rocket:

PR #205, opened 2015-11-17
Add a primo harvester

marktriggs:
The getty & MWDL mappings will depend on this.

### DIFF
--- a/lib/krikri/harvesters/primo_harvester.rb
+++ b/lib/krikri/harvesters/primo_harvester.rb
@@ -1,0 +1,106 @@
+module Krikri::Harvesters
+  ##
+  # A harvester implementation for Primo
+  #
+  # Accepts options passed as `:primo => opts`
+  #
+  # Options allowed are:
+  #
+  #   - bulk_size: The number of records to fetch from Primo per request
+  #     (default: 500)
+  #
+  class PrimoHarvester
+    include Krikri::Harvester
+
+    SEAR_NS = 'http://www.exlibrisgroup.com/xsd/jaguar/search'
+    PRIMO_NS = 'http://www.exlibrisgroup.com/xsd/primo/primo_nm_bib'
+
+    PrimoHarvestError = Class.new(StandardError)
+
+    def initialize(opts = {})
+      @opts = opts.fetch(:primo, {})
+      super
+
+      @opts[:bulk_size] ||= 500
+
+      @http_conn = Faraday.new do |conn|
+        conn.request :retry, :max => 3
+        conn.response :follow_redirects, :limit => 5
+        conn.response :logger, Rails.logger
+        conn.adapter :net_http
+      end
+    end
+
+    ##
+    # @return [Integer] the number of records available for harvesting.
+    def count
+      response = @http_conn.get(uri, :indx => '1', :bulkSize => '1')
+
+      unless response.status == 200
+        raise PrimoHarvestError, "Couldn't get record count"
+      end
+
+      total_hits = Nokogiri::XML(response.body)
+                   .xpath('//sear:DOCSET')
+                   .first
+                   .attr('TOTALHITS')
+
+      Integer(total_hits)
+    end
+
+    ##
+    # @return [Enumerator::Lazy] an enumerator of the records targeted by this
+    #   harvester.
+    def records
+      bulk_size = @opts.fetch(:bulk_size)
+
+      (1...count).step(bulk_size).lazy.flat_map do |offset|
+        response = @http_conn.get(uri, :indx => offset, :bulkSize => bulk_size)
+
+        unless response.status == 200
+          raise PrimoHarvestError, "Record fetch from #{offset} to " \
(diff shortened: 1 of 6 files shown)


marktriggs:
Do not merge!

I'll give this a once over soon--and write some tests for it.

markbreedlove:
Looks good!  Thanks for working on the tests.

The new `MarcXMLHarvester` that's in the works (https://github.com/dpla/KriKri/pull/204) uses a method named `#enumerate_records`, while this one has `#extract_records`.  Do you think those could be named the same to make them easier to recognize by someone who's looked at the other harvester already?

lib/krikri/harvesters/primo_harvester.rb
```diff
@@ -0,0 +1,106 @@
+module Krikri::Harvesters
+  ##
+  # A harvester implementation for Primo
+  #
+  # Accepts options passed as `:primo => opts`
+  #
+  # Options allowed are:
+  #
+  #   - bulk_size: The number of records to fetch from Primo per request
```
anarchivist: We'll need to have it accept additional Primo-specific options, which will vary across MWDL and Getty: `institution`, `loc`, `query`, and `query_exc` (see [the ingest details spreadsheet](https://docs.google.com/spreadsheets/d/1pACcWaevDNht7XbE_Sjs-Nnl4XWoQ2qqAeFaiBAxFMw/edit#gid=1534136574)). The first three are just string parameters, but `query_exc` should be able to take an array, which would lead to `query_exc` being added more than once to the request.

marktriggs:
@anarchivist At the moment the parameters are being set on a per-harvest basis just by providing them with the URI, like:

```
PrimoHarvester.new(:uri => 'http://primose1.getty.edu:1701/PrimoWebServices/xservice/search/brief?indx=1&institution=01GRI&loc=local,scope:(GETTY_OCP,GETTY_ROSETTA)&query=facet_local5,exact,DPLA')
```

and that's working OK for Getty and MWDL.  In fact, the only reason I pulled out `bulkSize` as special was because the harvester needs to override it in some cases, and I felt like making it a parameter made that a little more obvious.

Do you think it's worth modelling the other parameters explicitly in the harvester?  I'd assumed that different institutions could have quite different sets of base parameters, so opted for passing through whatever was provided by the caller without modification.

anarchivist:
This is just a case of me being thick, apparently. :stuck_out_tongue_closed_eyes: 

I'm of two minds on this. I think as is, it's probably fine, but I'd defer to @markbreedlove or @no-reply here. If they're good, I'd probably want us to add something in the documentation that the URI can/should include certain parameters to build out the query.

However, I'm a little curious about why this didn't build on [`Krikri::Harvesters::APIHarvester`](https://github.com/dpla/krikri/blob/develop/lib/krikri/harvesters/api_harvester.rb). This is the basis of Heidrun's [`MdlHarvester`](https://github.com/dpla/heidrun/blob/develop/app/harvesters/mdl_harvester.rb).

marktriggs:
Yeah, I'd be happy to discuss merging the Primo harvester with the API Harvester if that makes sense.  It wasn't really a conscious decision--the API Harvester seemed quite JSON-centric, and I'd been following along with the OAI harvester as my guide (which happens to not use the API Harvester either, but maybe it should?)

anarchivist:
To be fair, I think you're right that the `APIHarvester` as is is pretty JSON-centric. The `OAIHarvester`'s implementation is a little different because it's based on `ruby-oai`, which lovingly abstracts away some of the complexities like resumption tokens.

lib/krikri/parsers/primo_parser.rb
```diff
@@ -0,0 +1,29 @@
+module Krikri
+  ##
+  # A Primo parser. Uses XML parser with a root path to match the
+  # metadata path.
+  # @see Krikri::XmlParser
+  class PrimoParser < XmlParser
+    include Krikri::OaiParserHeaders
```
no-reply: Do we need OaiParserHeaders? If so, we should document why, and maybe change the name of that module to just `ParserHeaders` or something.

no-reply:
This looks good. Specs for `PrimoParser` would be a good idea. See also: https://github.com/dpla/KriKri/pull/205#discussion_r45349304

Otherwise :+1:

marktriggs:
Thanks y'all.  I've added a spec covering PrimoParser and dropped the OaiParserHeaders dependency.  I think I added it because I thought we were going to use it, but then we didn't :)

markbreedlove:
I'm having a hard time deciding what to do about those parameters -- whether to put them in the URI, or to pass them as `opts` properties. We should go ahead with this as it is now, and if we have some insight later, we can refactor it, since we will probably not be bound to it being one way or another until we try to build a UI around it.

:+1:

markbreedlove:
I'm fine with it being squashed and merged, if nobody else objects.

anarchivist:
Agreed w/ @markbreedlove. :+1:

marktriggs:
Squashed now.  Thanks all!

PR #206, opened 2015-11-19
Expose local_name through the Parser interface

marktriggs:
This is used for error reporting by Mapping::Error when a mapping error
occurs inside a ChildDeclaration block.

Without this, errors would throw:

```
NoMethodError: undefined method `local_name' for #<Krikri::PrimoParser:0x007fa42ec51500>
```

### DIFF
--- a/lib/krikri/parser.rb
+++ b/lib/krikri/parser.rb
@@ -29,6 +29,13 @@ module Krikri
     end
 
     ##
+    # @return [String] the local_name of the OriginalRecord wrapped by this
+    #   parser
+    def local_name
+      record.local_name
+    end
+
+    ##
     # A generic parser value.
     #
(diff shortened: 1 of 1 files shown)


no-reply:
:+1:

lib/krikri/parser.rb
```diff
@@ -29,6 +29,12 @@ def self.parse(record, *args)
     end
 
     ##
+    # Return the local_name value of the OriginalRecord wrapped by this parser.
```
no-reply: Maybe swap this for a YARD annotation?

marktriggs:
Comment fixed--thanks for the reminder.

markbreedlove:
Looks good to me, too.

PR #207, opened 2015-11-20
Add #invalidate! for LDP resources

no-reply:
Uses prov:invalidatedAtTime to mark resources as invalidated (soft
deleted). This is different from `#delete!` which sends an HTTP/LDP
DELETE request, permanently marking the resource as `410 Gone`.

We aim to invalidate, usually, rather than delete.

Implemented as a mixin that works with objects implementing an
`#rdf_source` method returning a `Krikri::LDP::RDFSource`.

### DIFF
--- a/lib/krikri/ldp/invalidatable.rb
+++ b/lib/krikri/ldp/invalidatable.rb
@@ -1,0 +1,107 @@
+module Krikri::LDP
+  ##
+  # Implements invalidation for `Krikri::LDP::Resource`s. This is different 
+  # from deletion, in that the resource continues to respond `200 OK`, and 
+  # return the representation, Nothing is removed from the LDP server.
+  # 
+  # Works as a mixin to `Krikri::LDP::Resource`, assuming an implementation of
+  # `#rdf_source`, which may simply return `self`.
+  #
+  # @example invalidating a resource
+  #   class MyResource
+  #     include Krikri::LDP::Resource
+  #     include Krikri::LDP::Invalidatable
+  #
+  #     def rdf_subject
+  #       @rdf_subject ||= RDF::URI('http://example.com/ldp/a/resource/path')
+  #     end
+  #   end
+  #  
+  #   invalidatable_resource = MyResource.new
+  #   # the resource must exist before it can be invalidated! 
+  #   invalidatable_resource.save
+  #
+  #   invalidatable_resource.invalidate!
+  #   invalidatable_resource.invalidated? # => true
+  #   invalidatable_resource.invalidated_at_time 
+  #   # => Thu, 03 Dec 2015 10:27:45 -0800
+  #
+  # @see http://www.w3.org/TR/2013/REC-prov-dm-20130430/#term-Invalidation 
+  #   for documentation on PROV invalidation
+  module Invalidatable
+    # @see RDF::PROV
+    INVALIDATED_BY_URI = RDF::PROV.wasInvalidatedBy
+    INVALIDATED_TIME_URI = RDF::PROV.invalidatedAtTime
+
+    ##
+    # Invalidates the resource by marking it with a `prov:invalidatedAtTime`. If
+    # an `RDF::Term` is passed as the first argument, that term is used as the 
+    # value of `prov:wasInvalidatedBy`.
+    #
+    # @example invalidating with an activity
+    #   invalidatable_resource.invalidate!(RDF::URI('http://example.org/moomin'))
+    #   invalidatable_resource.was_invalidated_by
+    #   # => #<RDF::URI:0x2acab846109c URI:http://example.org/moomin>
+    #
+    # @param activity_uri [RDF::Term] a URI for the invalidating activity. If 
+    #   none is given, this defaults to `nil` and no `prov:wasInvalidatedBy`
+    #   statement is added.
+    # @param ignore_invalid [Boolean] if true, supresses errors on already,
+    #   invalid records
+    #
+    # @raise [RuntimeError] when the resource does not exist or is already 
+    #   invalid; unless `ignore_invalid` is `true`
+    # @return [void]
+    def invalidate!(activity_uri = nil, ignore_invalid = false)
+      raise "Cannot invalidate #{rdf_subject}, does not exist." unless exists?
+
+      # force a reload unless we have cached an invalidatedAtTime
+      rdf_source.get({}, true) unless invalidated?
+      # we check invalidated again in case the reload came back invalid
+      if invalidated?
+        return if ignore_invalid
+        raise "Cannot invalidate #{rdf_subject}, already invalid." 
+      end
+
+      uri = RDF::URI(rdf_subject)
+      
+      rdf_source << [uri, INVALIDATED_BY_URI, activity_uri] unless 
+        activity_uri.nil?
+      rdf_source << [uri, INVALIDATED_TIME_URI, DateTime.now]
+
+      rdf_source.save
+    end
+
+    ##
+    # @return [Boolean] `true` if the resource has been marked invalidated.
+    def invalidated?
+      !invalidated_at_time.nil?
+    end
+
+    ##
+    # @return [DateTime, nil] the time this resource was marked invalidated;
+    #   gives `nil` if the resource has not been invalidated.
+    # 
+    # @note if two invalidatedAtTimes exist, we may get either of them back!
+    def invalidated_at_time
+      time = first_property(INVALIDATED_TIME_URI)
+      time.nil? ? nil : time.object
+    end
+
+    ##
+    # @return [RDF::URI, nil] the activity responsible for invalidating the 
+    #   resource
+    # 
+    # @note if two wasInvalidatedBys exist, we may get either of them back!
+    def was_invalidated_by
+      first_property(INVALIDATED_BY_URI)
+    end
+
+    private 
+    
+    def first_property(predicate)
+      res = rdf_source.query([RDF::URI(rdf_subject), predicate, nil])
+      res.empty? ? nil : res.first.object
+    end
+  end
+end
--- a/app/models/krikri/original_record.rb
+++ b/app/models/krikri/original_record.rb
@@ -3,6 +3,7 @@ module Krikri
   # Handles records as harvested, prior to mapping
   class OriginalRecord
     include Krikri::LDP::Resource
+    include Krikri::LDP::Invalidatable
 
     attr_accessor :content, :local_name, :rdf_subject
     attr_writer :content_type
--- a/lib/krikri/ldp.rb
+++ b/lib/krikri/ldp.rb
@@ -4,7 +4,8 @@ module Krikri
   # As LDP support develops, it might be possible to excract this or replace it
   # with a tool like the `ldp` gem.
   module LDP
-    autoload :Resource,     'krikri/ldp/resource'
-    autoload :RdfSource,    'krikri/ldp/rdf_source'
+    autoload :Resource,         'krikri/ldp/resource'
+    autoload :RdfSource,        'krikri/ldp/rdf_source'
+    autoload :Invalidatable,    'krikri/ldp/invalidatable'
   end
 end
--- a/lib/krikri/ldp/rdf_source.rb
+++ b/lib/krikri/ldp/rdf_source.rb
@@ -7,13 +7,26 @@ module Krikri::LDP
   module RdfSource
     extend ActiveSupport::Concern
     include Krikri::LDP::Resource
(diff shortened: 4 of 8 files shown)


no-reply:
The failures here don't look related to the changes, I restarted https://travis-ci.org/dpla/KriKri/builds/92053072 to see if `develop` might be failing, too.

anarchivist:
```
NameError: uninitialized constant Krikri::LDP::Invalidatable
/home/travis/build/dpla/KriKri/app/models/krikri/original_record.rb:6:in `<class:OriginalRecord>'
/home/travis/build/dpla/KriKri/app/models/krikri/original_record.rb:4:in `<module:Krikri>'
/home/travis/build/dpla/KriKri/app/models/krikri/original_record.rb:1:in `<top (required)>'
/home/travis/build/dpla/KriKri/lib/tasks/krikri.rake:10:in `<top (required)>'
/home/travis/.rvm/gems/ruby-2.1.3/gems/railties-4.1.14/lib/rails/engine.rb:654:in `block in run_tasks_blocks'
/home/travis/.rvm/gems/ruby-2.1.3/gems/railties-4.1.14/lib/rails/engine.rb:654:in `each'
/home/travis/.rvm/gems/ruby-2.1.3/gems/railties-4.1.14/lib/rails/engine.rb:654:in `run_tasks_blocks'
/home/travis/.rvm/gems/ruby-2.1.3/gems/railties-4.1.14/lib/rails/application.rb:399:in `block in run_tasks_blocks'
/home/travis/.rvm/gems/ruby-2.1.3/gems/railties-4.1.14/lib/rails/engine/railties.rb:13:in `each'
/home/travis/.rvm/gems/ruby-2.1.3/gems/railties-4.1.14/lib/rails/engine/railties.rb:13:in `each'
/home/travis/.rvm/gems/ruby-2.1.3/gems/railties-4.1.14/lib/rails/application.rb:399:in `run_tasks_blocks'
/home/travis/.rvm/gems/ruby-2.1.3/gems/railties-4.1.14/lib/rails/engine.rb:449:in `load_tasks'
/home/travis/build/dpla/KriKri/spec/internal/Rakefile:6:in `<top (required)>'
/home/travis/.rvm/gems/ruby-2.1.3/bin/ruby_executable_hooks:15:in `eval'
/home/travis/.rvm/gems/ruby-2.1.3/bin/ruby_executable_hooks:15:in `<main>'
```

lib/krikri/ldp/ivalidatable.rb
```diff
@@ -0,0 +1,6 @@
+module Krikri::LDP
```
anarchivist: Is this file necessary?

anarchivist:
I'm wondering if we need to add an `autoload` call in [`lib/krikri/ldp.rb`](https://github.com/dpla/KriKri/blob/feature/invalidate/lib/krikri/ldp.rb) to fix the build issue.

anarchivist:
Either that, or `require 'krikri/ldp/invalidatable'`.

anarchivist:
I think d1dc4d7 should fix the above build issue. @markbreedlove, I'm assigning this for review to you now.

anarchivist:
Yep, that fixed it. This has been rebased against `develop`, too.

markbreedlove:
It would help if this were added to the doc comment for the module:
> Implemented as a mixin that works with objects implementing an
> #rdf_source method returning a Krikri::LDP::RDFSource.

For the uninitiated, it can be unclear how `RDF::PROV.wasInvalidatedBy` and `RDF::PROV.invalidatedAtTime` come to be defined.  (At least, it's unclear to me.)

This is hard to evaluate without an example of how it would be used, but it seems clear enough how the statements are added in `#invalidate!`.

lib/krikri/ldp/invalidatable.rb
```diff
@@ -0,0 +1,63 @@
...
+        activity_uri.nil?
+      rdf_source << [uri, INVALIDATED_TIME_URI, DateTime.now]
+
+      rdf_source.save
+    end
+
+    ##
+    # @return [Boolean] `true` if the resource has been marked invalidated.
+    def invalidated?
+      !invalidated_at_time.nil?
```
anarchivist: Is it more important for the resource to have `prov:invalidatedAtTime` or `prov:wasInvalidatedBy`? Documentation here is probably beneficial as to why.

lib/krikri/ldp/invalidatable.rb
```diff
@@ -0,0 +1,63 @@
+module Krikri::LDP
+  ##
+  # Implements invalidatation for `Krikri::LDP::Resource`s. This is different
```
anarchivist: minor: typos this and next line ("invalidatation" and "deletetion")

lib/krikri/ldp/invalidatable.rb
```diff
@@ -0,0 +1,63 @@
...
+  # from deletetion, in that the resource continues to respond `200 OK`, and 
+  # return the representation, Nothing is removed from the LDP server.
+  module Invalidatable
+    INVALIDATED_BY_URI = RDF::PROV.wasInvalidatedBy
+    INVALIDATED_TIME_URI = RDF::PROV.invalidatedAtTime
+
+    ##
+    # Invalidates the resource by marking it with a `prov:invalidatedAtTime`.
+    #
+    # @return [void]
```
anarchivist: As @markbreedlove noted, an `@example` here would be useful.

lib/krikri/ldp/invalidatable.rb
anarchivist: FWIW, PROV-DM [seems to suggest](http://www.w3.org/TR/2013/REC-prov-dm-20130430/#term-Invalidation) that since Invalidation of an Entity occurs within the context of an Activity, it should probably be `prov:wasInvalidatedBy`.

no-reply:
I think the added documentation in the latest push should help clarify some of these questions.

lib/krikri/ldp/invalidatable.rb
no-reply: We're using `prov:invalidatedAtTime` as the canonical flag for invalidation; this *maybe* runs afoul of PROV---though I don't see any formal constraint that the activity's timestamp must match `invalidatedAtTime`---but it simplifies the logic a bit.

Especially, it means that we can use the same logic whether an activity is present or not; and that we don't need to reach into the activity to guarantee a timestamp.

lib/krikri/ldp/invalidatable.rb
```diff
@@ -0,0 +1,102 @@
...
+    # @raise [RuntimeError] when the resource does not exist or is already 
+    #   invalid
+    # @return [void]
+    def invalidate!(activity_uri = nil)
+      raise "Cannot invalidate #{rdf_subject}, does not exist." unless exists?
+
+      # force a reload unless we have cached an invalidatedAtTime
+      rdf_source.get({}, true) unless invalidated?
+      # we check invalidated again in case the reload came back invalid
+      raise "Cannot invalidate #{rdf_subject}, already invalid." if invalidated?
```
markbreedlove: Can you elaborate on why an exception is raised in this case?

Perhaps without realizing some subtleties to the usage of `!` I'd expect this to be idempotent in the fashion of this example:
```
irb(main):006:0> s = "A"
=> "A"
irb(main):007:0> s.downcase!
=> "a"
irb(main):008:0> s
=> "a"
irb(main):009:0> s = "a"
=> "a"
irb(main):010:0> s.downcase!  # <-- note this case
=> nil
irb(main):011:0> s
=> "a"
```
I'm thinking of how downcasing "a" does not result in an exception.  It returns `nil`, which I'm not sure is intuitive, but it doesn't cause an exception in a routine that might be processing items without wanting to check each one for whether it's already in the desired state. For example, I could see wanting to invalidate a number of records, having that fail partway through due to some random error elsewhere in the application, and then wanting to rerun the operation on the same list.

no-reply: I considered that part of the contract of `#invalidate!` would be to update the timestamp and activity. So we should be able to reliably expect something like this to be true:

```ruby
aggregation.invalidate!(activity_uri)
aggregation.invalidated_at_time.second == DateTime.now.second
aggregation.invalidated_by == activity_uri
```

But since we don't want to change the timestamp or activity if the resource was already invalid, we need to throw an error.

The other option is to allow the `#invalidate!` to "fail" silently, but that seemed likely to bite us, to me. Thoughts?

For your use case, something like `aggregation.invalidate! if aggregation.valid?` seems workable. In general, I think we would want to be working with valid records only, in most cases.

All this said, I think if we keep things as is, we need to give this a better error class than `RuntimeError`.

markbreedlove: I guess I see the point about the contract with the Activity to record _it_ as the invalidator and fail if that can't be done. I wish there could be some way to _warn_ the Activity that an aggregation was already invalidated, just because of the case where you might want to rerun an activity that invalidates records based on some query (without having to revise the query).  It seems that a special exception class is necessary, as you mentioned. Depending on the situation, the exception could be handled one way or another (e.g. logging and continuing).

Just out of curiosity, is there any possibility of having some kind of Activity in our system that's not one of the enqueued activities? Like a "manual run" activity?

markbreedlove:
Reassigning to @no-reply because he's going to ponder some things and probably make some updates.

no-reply:
@markbreedlove I force pushed a change on this that I think helps with the error condition issue. You can now do `#invalidate!(uri, true)` to ignore the request if it is already invalidated, rather than squawking. 

The default is still to raise an error when trying to invalidate already invalid records; client code can catch those errors and log as necessary.

While cooking this up, I originally wanted to use Ruby 2.0's named parameters. I've been using these more in other contexts lately, now that Ruby 1.9 support is officially deprecated, and maybe we should consider some internal guidelines for using them in our code. See https://robots.thoughtbot.com/ruby-2-keyword-arguments and some (stale?) talk about style guide additions at https://github.com/bbatsov/ruby-style-guide/issues/396. cc: @anarchivist @AudreyAltman

markbreedlove:
I'm in favor of that second parameter.

Also in favor of analyzing how how to refactor our function headers to use named parameters. I agree that we should do it deliberately, across the codebase, rather than having one odd method that follows a different pattern than the others.

I'm OK with this pull request.  @anarchivist @AudreyAltman, any observations, or should I merge this?

anarchivist:
I'm :+1: on this. @no-reply has addressed my concerns, and a reread of the PROV specs has helped. Adding some notes here for myself, so we can move this forward:

See [PROV-SEM on Invalidation](https://www.w3.org/TR/prov-sem/#invalidation) and [PROV-DM on Invalidation (as defined)](https://www.w3.org/TR/prov-dm/#concept-invalidation), most specifically:

> Invalidation<sup>[◊](https://www.w3.org/TR/2013/REC-prov-dm-20130430/#dfn-wasinvalidatedby)</sup>, written `wasInvalidatedBy(id; e, a, t, attrs)` in PROV-N, has:
>
> * `id`: an optional identifier for a invalidation;
> * `entity`: an identifier for the invalidated entity;
> * `activity`: an optional identifier for the activity that invalidated the entity;
> * `time`: an optional "invalidation time", the time at which the entity began to be invalidated;
> * `attributes`: an optional set of attribute-value pairs representing additional information about this invalidation.
>
> While each of `id`, `activity`, `time`, and `attributes` is **OPTIONAL**, at least one of them **MUST** be present.

See also [PROV-O's expanded terms](https://www.w3.org/TR/2013/REC-prov-o-20130430/#description-expanded-terms), and [example 2](https://www.w3.org/TR/2013/REC-prov-o-20130430/#narrative-example-expanded-1) immediately following: 

> The fourth category of expanded terms describes the lifetime of an Entity beyond being **generated** by an Activity and **used** by other Activities. For example, a painting could not have been displayed before it was painted, and it could not be sold after it was destroyed by fire. Similar to how Activities have start and end times, an Entity may be bound by points in time for which it was generated or is no longer usable. The properties `prov:generatedAtTime` and `prov:invalidatedAtTime` can be used to bound the starting and ending moments of an Entity's existence. The Activities that led to the generation or invalidation of an Entity can be provided using `prov:wasGeneratedBy` and `prov:wasInvalidatedBy`, respectively.

lib/krikri/ldp/invalidatable.rb
anarchivist: Closing the loop on this; I'm satisfied. @anarchivist: [my comment for documentation](https://github.com/dpla/KriKri/pull/207#issuecomment-170713212)

no-reply:
:rocket: @markbreedlove I think it's yours to :merge:

PR #208, opened 2015-12-18
Asynchronous harvesters

jambun:
Three new harvesters. We are submitting them in one pull request because they all use a new helper - async_uri_getter.rb which allows them to multithread GETs for items.

The harvesters are:
    NYPL
    Internet Archive
    Fedora (for UVa)

### DIFF
--- a/lib/krikri/async_uri_getter.rb
+++ b/lib/krikri/async_uri_getter.rb
@@ -1,0 +1,144 @@
+require 'faraday'
+require 'faraday_middleware'
+
+require 'net/http'
+require 'thread'
+require 'uri'
+
+module Krikri
+  ##
+  # Helper class for fetching multiple URLs concurrently. For example, to fetch
+  # 5 URLs in 5 threads:
+  #
+  #    urls = ['http://example.com/one',
+  #            'http://example.com/two',
+  #            'http://example.com/three',
+  #            'http://example.com/four',
+  #            'http://example.com/five']
+  #           .map { |url| URI.parse(url) }
+  #
+  #    getter = Krikri::AsyncUriGetter.new
+  #
+  #    requests = urls.map do |url|
+  #      getter.add_request(uri: url,
+  #                         opts: {
+  #                           follow_redirects: true
+  #                         })
+  #    end
+  #
+  # At this point, 5 threads are launched to fetch the list of URLs.  We can
+  # wait for them all to finish if we want to make sure we don't continue until
+  # all threads have terminated:
+  #
+  #    requests.map(&:join)
+  #
+  # Or simply access the responses and have our current thread block until
+  # they're available:
+  #
+  #    requests.each do |request|
+  #      request.with_response do |response|
+  #        if response.status == 200
+  #          puts "Response body: #{response.body}"
+  #        else
+  #          puts "Got return status: #{response.status}"
+  #        end
+  #      end
+  #    end
+  #
+  class AsyncUriGetter
+    MAX_REDIRECTS = 10
+
+    ##
+    # Create a new asynchronous URL fetcher.
+    #
+    # @param opts [Hash] a hash of the supported options, which are:
+    #
+    #  - follow_redirects: (true or false) -- whether to follow HTTP 3xx
+    #      redirects.
+    #
+    #  - max_redirects: [N] (default: 10) -- how many redirects to follow before
+    #      giving up
+    #
+    def initialize(opts: {})
+      @default_opts = { max_redirects: MAX_REDIRECTS }.merge(opts)
+    end
+
+    ##
+    # Run a request (in a new thread) and return a promise-like object for the
+    # response.
+    #
+    # @param uri [URI] the URI to be fetched
+    #
+    # @param headers [Hash<String, String>] HTTP headers to include with the
+    #   request
+    #
+    # @param opts [Hash] options to override the ones provided when
+    #   AsyncUriGetter was initialized. All supported options are available here
+    #   as well.
+    #
+    def add_request(uri: nil, headers: {}, opts: {})
+      fail ArgumentError, 'uri must be a URI' unless uri.is_a?(URI)
+      Request.new(uri, headers, @default_opts.merge(opts))
+    end
+
+    Request = Struct.new(:uri, :headers, :opts) do
+      def initialize(*)
+        super
+        @request_thread = start_request
+      end
+
+      ##
+      # Wait for the request thread to complete
+      #
+      def join
+        @request_thread.join
+      end
+
+      ##
+      # @yield [Faraday::Response] the response returned for the request
+      #
+      def with_response
+        yield(@request_thread.value)
+      end
+
+      private
+
+      ##
+      # Run the Faraday request in a new thread
+      #
+      def start_request
+        Thread.new do
+          http.get(uri) do |request|
+            headers.each do |header, value|
+              request.headers[header.to_s] = value
+            end
+          end
+        end
+      end
+
+      ##
+      # @return [Faraday::Connection] a connection with sensible defaults
+      #
+      def http
+        Faraday.new do |conn|
+          conn.request :retry,
+                       max: 4,
+                       interval: 0.025,
+                       interval_randomness: 0.5,
+                       backoff_factor: 2,
+                       exceptions: [Faraday::ConnectionFailed,
+                                    'Errno::ETIMEDOUT',
+                                    'Timeout::Error',
+                                    'Error::TimeoutError',
+                                    Faraday::TimeoutError]
+          if opts.fetch(:follow_redirects, false)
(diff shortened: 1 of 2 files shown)


lib/krikri/harvesters/async_uri_getter.rb
```diff
@@ -0,0 +1,74 @@
+require 'net/http'
+require 'thread'
+require 'uri'
+
+# Get URIs asynchronously
+class AsyncUriGetter
```
anarchivist: Should this class be namespaced somewhere under `Krikri`?

anarchivist: Also, the class and methods need documentation.

lib/krikri/harvesters/async_uri_getter.rb
```diff
@@ -0,0 +1,74 @@
...
+    end
+
+    private
+
+    def start_request
+      Thread.new do
+        response = nil
+        uri_to_try = uri
+        10.times do
+          http = Net::HTTP.new(uri_to_try.host, uri_to_try.port)
```
anarchivist: Could we decouple this from `Net::HTTP` by using Faraday?

lib/krikri/harvesters/ia_harvester.rb
```diff
@@ -0,0 +1,160 @@
...
+
+    ##
+    # Initialize, and set default options as appropriate.
+    #
+    # @param opts [Hash] a hash of options as defined by {.expected_opts}
+    #
+    # @example
+    #    Typical instantiation, good for most cases:
+    #
+    #      Krikri::Harvesters::IaHarvester.new(uri: 'http://archive.org/...')
```
anarchivist: What is the URI expected to be? A search URI? A collection URI?

lib/krikri/harvesters/ia_harvester.rb
```diff
@@ -0,0 +1,160 @@
...
+      identifiers.each do |id|
+        meta_uri = "#{DOWNLOAD_BASE_URI}/#{id}/#{id}_meta.xml"
+
+        batch << { :meta_request => @http.add_request(uri: URI.parse(meta_uri)),
+                   :id => id }
+      end
+
+      batch.each { |r| r[:meta_request].join }
+
+      batch.each do |record|
```
anarchivist: Can we add in a request to fetch the `*_files.xml` for the item as well? (see [here](https://github.com/dpla/ingestion/blob/develop/lib/fetchers/ia_fetcher.py#L271) for instance)

lib/krikri/harvesters/fedora_harvester.rb
```diff
@@ -0,0 +1,158 @@
+require 'uri'
+require 'async_uri_getter'
+
+module Krikri::Harvesters
+  ##
+  # A harvester implementation for Fedora 3
```
anarchivist: I think we should specify that this is specifically for UVA's Fedora 3 instance.

anarchivist: Does this belong in Heidrun instead?

lib/krikri/harvesters/async_uri_getter.rb
no-reply: :+1: to this. We also use Faraday in the OAI harvester: https://github.com/dpla/KriKri/blob/develop/lib/krikri/harvesters/oai_harvester.rb#L42

It would be nice to be able to have its benefits application-wide.

lib/krikri/harvesters/fedora_harvester.rb
marktriggs: We weren't totally clear about how harvesters should be split between Heidrun and KriKri, so this was something we'd intended to check.  I had the sense that generic harvesters lived in KriKri, while site-specific ones lived in Heidrun, so perhaps IA and NYPL should move to Heidrun as well?

For the sake of this PR we just kept them all together, but can easily move them around.

lib/krikri/harvesters/async_uri_getter.rb
marktriggs: Sounds good!  We left this a little sparse because we weren't sure if this was the direction to take with threading, but we'll switch it to Faraday, namespace & add documentation.

lib/krikri/harvesters/fedora_harvester.rb
anarchivist: Yeah, I think this makes sense to move IA and NYPL to Heidrun as well. Thanks.

marktriggs: Beauty.  I'll let James address the couple of issues raised that remain, and then we'll move them all over.

lib/krikri/harvesters/ia_harvester.rb
jambun: A search URI for a collection. I found this in:
https://github.com/dpla/ingestion/blob/develop/profiles/ia.pjs

An example I used while testing:
http://archive.org/advancedsearch.php?q=collection%3A%22bostonpubliclibrary%22&fl%5B%5D=identifier&output=json

lib/krikri/harvesters/ia_harvester.rb
jambun: Sure. I was originally going to get the _files.xml file, but it turned out to be unnecessary based on the mapping requirements. Should I attach it to the _meta.xml file somehow?

lib/krikri/harvesters/fedora_harvester.rb
jambun: My guess was it would be better to implement this as a generic Fedora 3 harvester. Happy to recast it as UVa specific if that's better.

lib/krikri/harvesters/ia_harvester.rb
anarchivist: :+1:

marktriggs:
This commit now just includes the Async HTTP request code.  All the action from before is now in Heidrun as PR https://github.com/dpla/heidrun/pull/38

anarchivist:
I'm generally :+1:, and I've asked @markbreedlove to take another glance, as I'm not sure how much this code on its own needs or would benefit testing.

markbreedlove:
I am trying to think of how exception-handling could be added and tested. It would be a good idea to have some tests included.

If one of the Faraday calls in `#http` results in an exception, we'll end up with a dead thread being assigned to `@request_thread`.  Later, when `#join` is called on that, in a harvester -- and I'm looking at `#enumerate_records` in `IaHarvester` in dpla/heidrun#38 -- the exception will bubble up. It appears that it's not handled either in `AsyncUriGetter` or `IaHarvester`, unless I'm overlooking something.

markbreedlove:
Would it be a good idea to call `#alive?` on the new Thread object and raise a specific exception in `#add_request` if it's not?

marktriggs:
Thanks for giving this some thought @markbreedlove.  It's a tricky area.

The semantics I had in mind were pretty much as you describe: if anything goes wrong with connecting to the host, or if the connection times out, it throws an exception inside of the request thread.  This bubbles to the initiating thread when it either calls `.join` or `.with_response`, and it can handle the exception at that point.

4xx and 5xx HTTP responses *don't* throw an exception, so you handle them in your `.with_response` block explicitly.  For these harvesters, we've used exceptions to signify unrecoverable events, so blowing up loudly is the goal.

I think the call to `http` is unlikely to fail in general because it's not actually making the connection at that point--just setting up Faraday with the appropriate middleware.  But if it did fail I think it would still be handled in the same way.

I don't think we can rely on `Thread#alive?` to do what we want because the timing can make it unpredictable.  For example, even if you had:

     mythread = Thread.new do
       raise "this thread is definitely going to fail"
     end

     unless mythread.alive?
       raise "Something went wrong with the thread"
     end

Whether or not the `mythread.alive?` call succeeds depends on what the OS scheduler did.  If it didn't give that new thread any time, then it'll still think things are fine when the check runs, even though the thread is actually doomed.

Maybe it's worth writing some tests for all of this, just to assert what happens when one of these fatal errors takes place?  I realize now that I'd been relying on the test suites of the harvesters to implicitly test this code, but now that they're moved to Heidrun I should probably add a new spec file anyway.  What do you think?

anarchivist:
> Maybe it's worth writing some tests for all of this, just to assert what happens when one of these fatal errors takes place? I realize now that I'd been relying on the test suites of the harvesters to implicitly test this code, but now that they're moved to Heidrun I should probably add a new spec file anyway. What do you think?

:+1: I think this is a great idea. Thanks, @marktriggs!

markbreedlove:
I think something small to assert what happens when the fatal exceptions takes place would be useful. It would be good to make it apparent to everyone that it works as you've described.

marktriggs:
Especially me :)

no-reply:
I have some minor changes to add. Since this is in a `hudmol` branch, I can't push directly to the PR. I'm closing it and opening another with my changes added.

PR #209, opened 2015-12-22
Fix coordinates for MAP 3.1 JSON Crosswalk

no-reply:
The 3.1 crosswalk mistakenly mapped arrays of lats & long into a string, resulting in coordinates like `"[\"12.34\"], [\"56.78\"]"`. This patch grabs the first value of each, resulting in the correct value.


anarchivist:
:100:

PR #211, opened 2016-01-12
Fix failing ProdSearchIndex spec that was skipped in CI

anarchivist:
This reorders the spec to used the :opts passed by RSpec to be only used in the context describing `#initialize`, and removes the mock that was preventing the appropriate method from firing. It also modifies the `entities query` shared context to set subject URI for the built aggregations.

The specs that test Krikri::ProdSearchIndex with an Elasticsearch connection were being skipped in Travis, and they failed for me locally. It looked in some cases as though it was trying to make a connection to http://example.org/moomin-index/moomin as used in the describe block.

### DIFF
--- a/lib/krikri/engine.rb
+++ b/lib/krikri/engine.rb
@@ -83,6 +83,12 @@ module Krikri
     end
 
     initializer :aggregation do
+      class NamespaceError < RuntimeError
+        def initialize(uri)
+          super("Tried to get DPLA ID for non-DPLA URI #{uri}")
+        end
+      end
+
       DPLA::MAP::Aggregation.class_eval do
         include Krikri::MapCrosswalk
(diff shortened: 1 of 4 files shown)


no-reply:
(assigning to self; I have a few changes on the way)

no-reply:
Some updates have been pushed that improve the test coverage--actually testing that the correct ids are returned.

I'd propose rebasing so be884db is the first commit, and the other two can be squashed together.

anarchivist:
@no-reply :+1: On it. Once it's rebased and squashed I'll tag someone else for review.

anarchivist:
Rebased and :squash:ed.

markbreedlove:
This looks fine to me. Ready to merge?

no-reply:
:merge: :+1:

PR #212, opened 2016-01-13
Asynchronous harvesters

no-reply:
Replaces #208. See discussion there.

4e272bd is the only commit changed from the previous discussion. It contains some YARD annotation updates, minor style changes, and **one substantial change**:

  - `#http` is cached within the `Request` instance. This should make no difference now (the connection and instance would have garbage collected at the same time anyway), but will prevent spawning multiple `Faraday` connections per `Request` if for some reason `Resquest` changes in the future. https://github.com/dpla/KriKri/compare/asynchronous-harvesters?expand=1#diff-a93afa836b08961677e7a6870c08c992R106

@anarchivist and/or @markbreedlove should sanity check that before merging.

### DIFF
--- a/lib/krikri/async_uri_getter.rb
+++ b/lib/krikri/async_uri_getter.rb
@@ -7,9 +7,9 @@ require 'uri'
 
 module Krikri
   ##
-  # Helper class for fetching multiple URLs concurrently. For example, to fetch
-  # 5 URLs in 5 threads:
+  # Helper class for fetching multiple URLs concurrently. 
   #
+  # @example to fetch 5 URLs in 5 threads
   #    urls = ['http://example.com/one',
   #            'http://example.com/two',
   #            'http://example.com/three',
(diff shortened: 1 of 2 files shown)


no-reply:
We should probably squash the first two commits, here, but keep my changes separate.

anarchivist:
I'm :+1:. I'm going to ping @jambun and @marktriggs here since this is a slight tweak of their initial PR.

marktriggs:
My only slight qualm here is that the caching change will cause a single Faraday instance to be shared by multiple threads, so we'll be reliant on Faraday being thread-safe.  A bit of Googling suggests that it probably is (https://github.com/lostisland/faraday/issues/370) but that it's a sort of "best effort" thread safety at the moment.

no-reply:
@marktriggs maybe I'm reading this wrong, but it looks to me like there's only ever one thread (at a time) associated with each `Request` instance. Is that wrong?

marktriggs:
Oh sorry, no, you're quite right.  I think an earlier version of this code split the threads from the requests in a slightly different way, but I was getting mixed up.

I think it's fine then!  :+1: from me :)

no-reply:
:merge: :grey_question:

no-reply:
:laughing:

PR #213, opened 2016-01-14
Limit `RDF::URI.cache` capacity

no-reply:
By default, `RDF::URI::CACHE_SIZE` is `-1` and all URIs initialized with
`RDF::URI.intern` are cached for the life of the application. In job
scenarios, this leads to a slow growth of the cache, as more unique URIs
are added. For a good number of the URIs we're using (e.g. LDP URIs and
their `#sourceResource` fragments), use is scoped to the processing of
one record. Limiting the cache size allows these to cycle out, while
commonly used URIs should stay in the cache avoiding duplicate objects
in memory.

The cache size is given by `Krikri::Settings['uri_cache_size']`, and
defaults to `1_000_000` (cached URIs). We could choose to change the
cache size on various worker boxes by installing unique settings
values.

Setting the cache size at application startup causes a warning to be
thrown that the previous value of the "constant" `CACHE_SIZE` is being
overwritten. An issue is open to find a way to squash the warnings,
hopefully with a better cache capacity management pattern, at
https://github.com/ruby-rdf/rdf/issues/256

### DIFF
--- a/app/models/krikri/original_record.rb
+++ b/app/models/krikri/original_record.rb
@@ -81,12 +81,24 @@ module Krikri
         record
       end
 
+      ##
+      # @return [String] the URI namespace/LDP container for resources of this 
+      #   class
       def base_uri
         Krikri::Settings['marmotta']['record_container']
       end
 
+      ##
+      # @param localname [#to_s] the unique portion of the URI to build; this 
+      #   will be combined with `.base_uri` to form the returned URI
+      # @return [RDF::URI] a URI built from the `.base_uri` and the `local_name` 
+      #   parameter
+      #
+      # @note we use `RDF::URI.intern` (rather than `.new`) to avoid allocating
+      #   memory for a new object for the LDP base URI, taking advantage of 
+      #   `RDF::URI.cache`.
       def build_uri(local_name)
-        RDF::URI(base_uri) / local_name
+        RDF::URI.intern(base_uri) / local_name
       end
 
       ##
--- a/lib/krikri/engine.rb
+++ b/lib/krikri/engine.rb
@@ -57,6 +57,11 @@ module Krikri
       Krikri::Settings = Kernel.const_get(RailsConfig.const_name)
     end
 
+    initializer :uri_cache do
+      RDF::URI::CACHE_SIZE = 
+        (Krikri::Settings['uri_cache_size'] || 1_000_000).to_i.freeze
+    end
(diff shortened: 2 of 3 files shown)


app/models/krikri/original_record.rb
```diff
@@ -86,7 +86,7 @@ def base_uri
       end
 
       def build_uri(local_name)
-        RDF::URI(base_uri) / local_name
+        RDF::URI.intern(base_uri) / local_name
```
anarchivist: A comment here referencing the reason for the change would be useful

anarchivist:
Looks good, other than a minor nit about adding some documentation. Are we going to be impacted by this other places in the ingestion workflow (e.g. on mapping or enrichment)? Passing to @markbreedlove for a second look.

no-reply:
Does the force push clarify?

`RDF::URI.intern` uses the cache, avoiding allocating (and subsequently GCing) new memory for the URI. In this case, the URI is the LDP original records container, and will be used for all of the processed records, so we prefer that it hits the cache.

no-reply:
> Are we going to be impacted by this other places in the ingestion workflow (e.g. on mapping or enrichment)? 

Yeah. I think this validates what @markbreedlove said originally: this memory pattern is likely playing out in mappings, too. Limiting the cache capacity should take care of it there, same as here.

Generally, since we're working with small graphs, almost always one at a time, a small-ish cache size ought to do us good. `1_000_000` is probably vastly bigger than we need, but the memory cost is small enough that it seems like a sane default.

anarchivist:
> Does the force push clarify?

For me, yep. :+1:

markbreedlove:
I've had a look at this and the relevant parts of `RDF::URI`, and I think this is good!

@no-reply, do you know where `RDF::URI.intern` is called in the mapping, or is that a separate PR for later?

markbreedlove:
Actually, it seems that setting `RDF::URI::CACHE_SIZE` the way you do in `lib/krikri/engine.rb` should take care of the mapper, too. Is that correct?

no-reply:
In both cases, the main place `.intern` is being called where it might be undesirable is `RDF::URI#/`.  

The cache size limit should take care of this problem generally, and it should be reasonably safe to use `.intern`, with the worst case cost being that the handful of frequently recurring URIs fall out of the cache periodically, then are readded on next use.

PR #214, opened 2016-01-27
Allow OAI harvesters to use alternate ID sources

no-reply:
By default, we use the `identifier` field of the OAI header to seed the
record id minter. In the legacy system, the ids used were derived from
the `dc:identifier` field in `metadata`.

This patch allows harvesters for those sources to select a field with an
xpath expression, passed as `[:oai][:id_path]` in options.

### DIFF
--- a/lib/krikri/harvesters/oai_harvester.rb
+++ b/lib/krikri/harvesters/oai_harvester.rb
@@ -22,6 +22,8 @@ module Krikri::Harvesters
   #   - metadata_prefix: A string specifying the metadata prefix. e.g. 'oai_dc'.
   #   - from: The begin date for the harvest.
   #   - until: The end date for the harvest.
+  #   - id_path: An alternate xpath (e.g. '//dc:identifier') from which to load
+  #              raw identifier strings. Default:  <record><header><identifier>.
   #
   # @see http://www.rubydoc.info/gems/oai/OAI/Client
   class OAIHarvester
@@ -32,12 +34,16 @@ module Krikri::Harvesters
     # @param opts [Hash] options to pass through to client requests.
     #   Allowable options are specified in OAI::Const::Verbs. Currently :from,
     #   :until, :set, and :metadata_prefix.
+    #   Additionally, you may pass an xpath string to `:id_path` specifying the
+    #   location of the IDs.
     # @see OAI::Client
     # @see #expected_opts
     def initialize(opts = {})
       opts[:harvest_behavior] ||= OAISkipDeletedBehavior
       super
-      @opts = opts.fetch(:oai, {})
+
+      @opts    = opts.fetch(:oai, {})
+      @id_path = @opts.delete(:id_path) { false }
 
       http_conn = Faraday.new do |conn|
         conn.request :retry, :max => 3
@@ -86,8 +92,12 @@ module Krikri::Harvesters
       opts = @opts.merge(opts)
       request_with_sets(opts) do |set_opts|
         client.list_records(set_opts).full.lazy.flat_map do |rec|
-          @record_class.build(mint_id(rec.header.identifier),
-                              record_xml(rec))
+          begin
+            @record_class.build(mint_id(get_identifier(rec)),
+                                record_xml(rec))
+          rescue => e
+            Krikri::Logger.log(:error, e.message)
+          end
         end
       end
     end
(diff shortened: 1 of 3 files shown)


no-reply:
This is a blocker for the SCDL reharvest.

spec/lib/krikri/harvesters/oai_harvester_spec.rb
```diff
@@ -453,6 +453,55 @@
...
+          subject.records.first
+        end
+
+        context 'and bad path' do
+          let(:args) do
+            { uri: 'http://example.org/endpoint', 
+              oai: { id_path: '//dc:not_a_path' } }
+          end
+          
+          it 'falls back on the header identifier' do
```
no-reply: Is this acceptable? I don't believe the legacy ingestion system has a fallback behavior, but since `dc:identifier` is not guaranteed to be present across records/collections, it seems like we would not want to fail on any records that are missing the field.

anarchivist: I think this is a question for @guegueng. My first impression is that this should be construed as a fatal error for the given record, as it indicates a pretty serious data quality or consistency issue with the provider. Maybe we can chat about it for a moment on today's checkin?

no-reply: We agreed to allow these identifiers to fail the records. We will capture the OAI id and raise an error so the record can be logged when skipped.

no-reply:
The behavior is changed to fail and log missing IDs.

@anarchivist take another look? I pre-squashed and force pushed, so if it looks good, it's ready to go from my perspective.

anarchivist:
How do exceptions get handled in enqueued harvester jobs?

no-reply:
Hmmmm... I was thinking that exceptions in this scope would be caught and logged at https://github.com/dpla/KriKri/blob/develop/lib/krikri/harvester.rb#L132-L144, but I'm less sure now and want to double check.

no-reply:
Well, the error handling sent me on a wild goose chase. It led me here: https://gist.github.com/no-reply/4b38f26b3fe32ad266a7

The take away is that `#records` implementations for individual harvesters need to catch and log errors individually. There's really not anything we can do in the base class to handle this uniformly, as far as I can see.

no-reply:
In the meanwhile, a fix for OAI is up.

lib/krikri/harvesters/oai_harvester.rb
```diff
@@ -207,5 +218,31 @@ def record_xml(rec)
         .add_namespace_definition(nil, 'http://www.openarchives.org/OAI/2.0/')
       doc.to_xml
     end
+
+    private
+
+    def get_identifier(record)
+      return record.header.identifier unless @id_path
+
+      xml = Nokogiri.XML(record_xml(record))
```
markbreedlove: I notice that we have to call `Nokogiri.XML` here because the return value of `record_xml` is a string, via its `doc.to_xml`.  But `Nokokiri.XML` is already called in `record_xml` in https://github.com/dpla/KriKri/blob/feature/oai-id-selection/lib/krikri/harvesters/oai_harvester.rb#L216 ... Would it be worth it to refactor so that the unserialization doesn't have to happen twice?

markbreedlove:
My only comment is regarding that `Nokogiri.XML`.  This looks good and makes sense to me.

I have to trust that the exception handling in `#get_identifier` is going to handle the kinds of exceptions that `xml.xpath` is going to throw -- I'm not too familiar with it.

lib/krikri/harvesters/oai_harvester.rb
no-reply: Yeah, I thought about that a bit, but didn't come up with anything that seemed much better.

I guess we could add an `as_doc: false` option to `#record_xml` so it can return the altered Nokogiri document.  Does that sound like a good fix?

no-reply:
> I have to trust that the exception handling in #get_identifier is going to handle the kinds of exceptions that xml.xpath is going to throw -- I'm not too familiar with it.

Other exceptions are probably possible, but the handling in `#records` should catch them (without the specialized error reporting).

lib/krikri/harvesters/oai_harvester.rb
markbreedlove: Sure, that sounds best.

markbreedlove:
OK, I'm fine with the exception handling. Other kinds of XML exceptions would evidently be due to XML syntax problems outside of the scope of the ID. If you want to put in the `as_doc: false` change, I'm happy to merge this if nobody else has other observations.

no-reply:
I think this is ready, at long last. :package: :shipit: :question:

PR #215, opened 2016-02-03
Filter invalid records from `#entities` and `#entity_ids`

markbreedlove:
By default, filter aggregations from `Krikri::Activity#entity_uris` and `#entities` that have been invalidated -- i.e. soft-deleted.

See https://www.w3.org/TR/prov-o/#invalidatedAtTime
See https://github.com/dpla/KriKri/commit/aab20d906e67f23aa5728bec3c6048e081dfd016

The most significant change is to `Krikri::ProvenanceQueryClient`, which is responsible for the SPARQL query that either includes or excludes invalidated aggregations.

Addresses https://issues.dp.la/issues/8237

If there's any concern about making the SPARQL query more complex and having that negatively impact query speed, I've benchmarked the old and new queries; and the new one, without invalidated records, appears to complete faster.  I've benchmarked these by restarting PostgreSQL and Marmotta in between, and by rebooting server VMs in my development environment, to avoid having caching skew the results.

### DIFF
--- a/app/models/krikri/activity.rb
+++ b/app/models/krikri/activity.rb
@@ -105,10 +105,18 @@ module Krikri
     # Return an Enumerator of URI strings of entities (e.g. aggregations or
     # original records) that pertain to this activity
     #
+    # @param  include_invalidated [Boolean] Whether to include entities that
+    #   have been invalidated with prov:invalidatedAtTime. Default: false
+    #
     # @return [Enumerator] URI strings
-    def entity_uris
+    #
+    # @see Krikri::ProvenanceQueryClient#find_by_activity regarding
+    #   invalidation.
+    #
+    def entity_uris(include_invalidated = false)
       activity_uri = RDF::URI(rdf_subject)  # This activity's LDP URI
-      query = Krikri::ProvenanceQueryClient.find_by_activity(activity_uri)
+      query = Krikri::ProvenanceQueryClient
+        .find_by_activity(activity_uri, include_invalidated)
       query.each_solution.lazy.map do |s|
         s.record.to_s
       end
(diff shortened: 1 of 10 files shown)


app/models/krikri/activity.rb
```diff
@@ -105,10 +105,20 @@ def to_s
     # Return an Enumerator of URI strings of entities (e.g. aggregations or
     # original records) that pertain to this activity
     #
+    # @param  include_invalidated [Boolean] Whether to include entities that
+    #                                       have been invalidated with
```
no-reply: can we indent these less aggressively? It seems like two spaces on hanging lines should be enough.

no-reply:
This looks good to me, apart from my request re: YARD formatting.

It occurred to me while reading it that we could refactor to make `ProvenanceQueryClient` a bit more flexible. I'm thinking along the lines of:

```ruby
q = ProvenanceQueryClient.find_by_activity(activity_id)
q.filter_invalidated.execute
```

We'd need to look at the `RDF::Query` interface, but I think this would work nicely if we make `ProvenanceQueryClient` a `Query` and support method chaining on it.

I'm :+1: this as is; the above is probably out of scope for the original ticket. I just wanted to note this so we could get the conversation moving.

no-reply:
@markbreedlove :squash: :question: 

(assuming you aren't dying to take on the suggested refactor)

markbreedlove:
Thanks for your review, @no-reply!

I've fixed that whitespace and squashed. I agree on the refactor, assuming we'd subclass `RDF::SPARQL::Query` but I think as I believe you do that it'd be best to pick it up in an upcoming iteration.

PR #216, opened 2016-02-10
Allow use of RSpec 3.4.x

no-reply:
RSpec is nearing a 3.5 release, which will be necessary for Rails 5 compatibility. This gets us closer to that.

It also fixes up warnings raised by newer versions of RSpec.

The dependency is loosened so engine users aren't forced to upgrade.

### DIFF
--- a/app/models/krikri/original_record.rb
+++ b/app/models/krikri/original_record.rb
@@ -30,8 +30,9 @@ module Krikri
       # @param identifier [#to_s] a string representing the #local_name or
       #   fully qualified URI for the resource.  Identifier may have a mime type
       #   extension, ie. '123.xml'.
+      #
       # @return [OriginalRecord] the instantiated record.
-      # @raise when no matching record is found in the LDP datastore
+      # @raise [LoadError] when no matching record is found in the LDP datastore
       def load(identifier)
         identifier = identifier.to_s.split('/').last if
           identifier.start_with? base_uri
@@ -42,8 +43,7 @@ module Krikri
           record = new(identifier)
         end
 
-        raise NameError, "No #{self} found with id: #{identifier}" unless 
-          record.exists?
+        raise LoadError.new(identifier), 'No record found' unless record.exists?
 
         if identifier.include?('.')
           record.rdf_subject = "#{base_uri}/#{identifier}"
@@ -201,6 +201,24 @@ module Krikri
     #   requests.
     def headers
       { 'Content-Type' => content_type }
+    end
+
+    ##
+    # Error class expressing failed load of an OriginalRecord 
+    class LoadError < RuntimeError
+      ##
+      # @param identifier [#to_s]
+      def initialize(identifier)
+        @identifier = identifier
+      end
+      
+      ##
+      # Augments the raised message with generic failure information
+      #
+      # @see Exception#message
+      def message
+        "Could not load OriginalRecord with `#{@identifier}: #{super}"
+      end  
     end
   end
 end
--- a/app/models/krikri/original_record.rb
+++ b/app/models/krikri/original_record.rb
@@ -42,7 +42,8 @@ module Krikri
           record = new(identifier)
         end
 
-        raise "No #{self} found with id: #{identifier}" unless record.exists?
+        raise NameError, "No #{self} found with id: #{identifier}" unless 
+          record.exists?
 
         if identifier.include?('.')
           record.rdf_subject = "#{base_uri}/#{identifier}"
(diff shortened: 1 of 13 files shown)


app/models/krikri/original_record.rb
```diff
@@ -42,7 +42,8 @@ def load(identifier)
           record = new(identifier)
         end
 
-        raise "No #{self} found with id: #{identifier}" unless record.exists?
+        raise NameError, "No #{self} found with id: #{identifier}" unless
```
no-reply: I'm not 100% sure about `NameError`, here. See: http://ruby-doc.org/core-2.2.0/NameError.html

Open to other suggestions, but it seems important to have an error other than `RuntimeError` for this.

anarchivist: [`Krikri::OriginalRecord#exists?`](https://github.com/dpla/KriKri/blob/feature/rspec-3.4/lib/krikri/ldp/resource.rb#L91-L103) is catching `Faraday::ResourceNotFound`. Is something like that preferable?

anarchivist: Looking at this again, maybe something like `Krikri::LDP::Resource::NotFoundError` if you want to keep it scoped to the module...? (That feels a little unwieldy, but ... ¯\_(ツ)_/¯)

no-reply: That feels awkward to me, too. I'm not sure `NotFound` is what we're looking for, here.

Probably we want our own error type... but that raises the question: where does the custom error sit in the hierarchy? Underneath `NameError`.

This line of thinking is why I pushed it up and asked for reviewer input. :smile:

anarchivist:
Do the remaining specs that aren't checking for instances of specific error classes indicate that we should be either testing for a specific error class?

no-reply:
RSpec's opinion is that a `raise_error` matcher should get either an error class or a message matcher (or both). In some cases, I chose to simply match the text, as `RuntimeError` seemed fine.

I think this clears all the warnings raised by the upgrade.

anarchivist:
Got it. Thanks!

markbreedlove:
@no-reply can you say how the new version of Rspec changes the exception handling?

Also, what was the Rspec change that lead to https://github.com/dpla/KriKri/pull/216/files#diff-a2cb539e5da95cc160b736d98af8443bR10 ?

no-reply:
@markbreedlove RSpec is now (correctly) opinionated that the `raise_error` shouldn't be used without giving some *specific* error, either by class or by matching message. This avoids passing in cases where, e.g., you're expecting some kind of `RuntimeError`, but in fact `NoMethodError` is raised.

I'll need to check out the branch again to recall what's happening with `contain_exactly`. I think it's something mostly unrelated that changed, causing the shared matcher to fail because it was testing the wrong thing. `include` is the correct test for "have provided_label".

I'll take a peek and make sure this change doesn't belong in its own commit/PR.

no-reply:
Following up on this: it seems like `contain_exactly` changed subtly in [RSpec 3.3](http://rspec.info/blog/2015/06/rspec-3-3-has-been-released/). This ran afoul of the pattern matching we're using in the shared example:

`actual.query(pattern).map(&:object)` yields an `RDF::Literal` instance, rather than the exact object. The `include` matcher still checks this in a way that takes advantage of RDF.rb's value-space literal equality.

As noted above, `include` is more precise for the test set, in either case. If we wanted to do `contain_exactly` in another context, we could cast the results.

markbreedlove:
Thanks, @no-reply!  That sounds good to me.

Like you, I have my issues with `NameError`, and I'm in favor of creating our own exception for this purpose. If `Krikri::LDP::Resource::NotFoundError` is too deeply nested, we may entertain the idea of a `Krikri::Error` module, and putting it there, if it's the kind of exception that we might want to handle throughout the application. It might not be called `NotFoundError` but something like `ResourceNotFoundError` to be more specific without requiring it to be too deeply nested. That's assuming it would have relevance in many "resource not found" situations. That could be finessed with a good choice of naming.

Apart from that, I have no further observations and think this is fine to merge.

markbreedlove:
I'll just assign this back to you, so you can route it to @anarchivist or me when you've finished making changes.

no-reply:
> I'll just assign this back to you, so you can route it to @anarchivist or me when you've finished making changes.

That sounds good, @markbreedlove.

WRT `LDP::NotFound` vs. `NameError`; my main concern is with `NotFound` is that it may be too blunt. It is really intended to represent 404 response codes, and simulating one seems like it might not be the best thing for client handling. The more I think about it, the more I think a custom error for this case is called for here.

What do you think of `Krikri::OriginalRecord::LoadError` ?

markbreedlove:
`Krikri::OriginalRecord::LoadError` looks good. `OriginalRecord` is a class, not a module ... is that OK?

no-reply:
Yeah, should be fine.

no-reply:
I think this is ready to go.

anarchivist:
:+1:

PR #218, opened 2016-02-22
Call `next` when logging errors in `#record`

no-reply:
The OAIHarvester previously gave `true` in `#records` when an error was caught building the record. `#run` handles `nil`, so we should return that instead.

Logging an error returns `true`, so code like this:

```ruby
map do |r|
  begin
    some_logic(r)
  rescue => e
    log(e)
  end
end
```

...results in `true` in the mapped enum. Calling `next` gives `nil` as required.

This fixes an issue originally identified here: https://dpla.slack.com/archives/tech/p1456162536000069


anarchivist:
:+1: Anyone else want to take a look?

PR #221, opened 2016-02-25
Entity Consumer Refactor/Deprecation

no-reply:
Refactors `EntityConsumer` to respond to `#entities` and to accept a generalized entity_source.  For the moment, it retains the previous generator activity behavior as a deprecated interface, to continue to serve the index classes, which are tightly coupled to `Activity`. `Mapper::Agent` and `Enricher` are refactored to use the new interface.

Moves activity URI lookup to `Activity` & exposes more public interfaces for working with activity URIs; Refactors `EntityConsumer` to rely on this logic, instead of using a coarse string based matcher.

Tests and docs are fixed up throughout.

### DIFF
--- a/app/models/krikri/activity.rb
+++ b/app/models/krikri/activity.rb
@@ -10,15 +10,40 @@ module Krikri
   class Activity < ActiveRecord::Base
     # @!attribute agent
     #    @return [String] a string representing the Krikri::SoftwareAgent
-    #                     responsible for the activity.
+    #      responsible for the activity.
+    # @!attribute start_time
+    #    @return [DateTime] a datestamp marking the activity's start
     # @!attribute end_time
     #    @return [DateTime] a datestamp marking the activity's competion
     # @!attribute opts
     #    @return [JSON] the options to pass to the #agent class when running
-    #                   the activity
-    # @!attribute start_time
-    #    @return [DateTime] a datestamp marking the activity's start
+    #      the activity
+    
     validate :agent_must_be_a_software_agent
+
+    ##
+    # @example building a valid URI from the base
+    #   Krikri::Activity.base_uri / 1
+    #   
+    # @return [RDF::URI] the configured base URI for this class
+    def self.base_uri
+      RDF::URI.intern(Krikri::Settings['marmotta']['ldp']) /
+        Krikri::Settings['prov']['activity']
+    end
+
+    ##
+    # @param uri [#to_s] a uri for this activity
+    #
+    # @return [Krikri::Activity] the activity with the given uri
+    #
+    # @raise [RuntimeError] if the URI form does not match the activity
+    # @raise [ActiveRecord::RecordNotFound] if no activity is found
+    def self.from_uri(uri)
+      raise "Cannot find #{self} from URI: #{uri}" unless 
+        uri.start_with? base_uri
+
+      find(uri.to_s.sub(base_uri.to_s, '').sub('/', ''))
+    end
 
     def agent_must_be_a_software_agent
       errors.add(:agent, 'does not represent a SoftwareAgent') unless
@@ -81,7 +106,9 @@ module Krikri
     def agent_instance
       @agent_instance ||= agent.constantize.new(parsed_opts)
     end
-
+    
+    ##
+    # @return [Hash] the options parsed as JSON
     def parsed_opts
       JSON.parse(opts, symbolize_names: true)
     end
@@ -89,11 +116,9 @@ module Krikri
     ##
     # @return [RDF::URI] the uri for this activity
     def rdf_subject
-      RDF::URI(Krikri::Settings['marmotta']['ldp']) /
-        Krikri::Settings['prov']['activity'] / id.to_s
+      self.class.base_uri / id.to_s
     end
     alias_method :to_term, :rdf_subject
-    
 
     ##
     # @return [String] a string reprerestation of the activity
--- a/app/models/krikri/activity.rb
+++ b/app/models/krikri/activity.rb
@@ -39,7 +39,8 @@ module Krikri
     # @raise [RuntimeError] if the URI form does not match the activity
     # @raise [ActiveRecord::RecordNotFound] if no activity is found
     def self.from_uri(uri)
-      raise "Cannot find #{self} from URI: #{uri}" unless 
+      raise "Cannot find #{self} from URI: #{uri}; " \
+            "the requested uri does not match #{base_uri}" unless 
         uri.start_with? base_uri
 
       find(uri.to_s.sub(base_uri.to_s, '').sub('/', ''))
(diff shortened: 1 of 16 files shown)


app/models/krikri/activity.rb
```diff
@@ -10,16 +10,41 @@ module Krikri
...
+    
     validate :agent_must_be_a_software_agent
 
+    ##
+    # @example building a valid URI from the base
+    #   Krikri::Activity.base_uri / 1
+    #   
+    # @return [RDF::URI] the configured base URI for this class
+    def self.base_uri
+      RDF::URI(Krikri::Settings['marmotta']['ldp']) /
```
anarchivist: Do we want an interned instance here?

no-reply: Seems like, yes. Otherwise we are instantiating a new one each time `.base_uri` is called.

lib/krikri/entity_behavior.rb
```diff
@@ -1,19 +1,36 @@
...
   # @see Krikri::Activity#entities
   # @see lib/krikri/entity_behaviors
-  #
   class EntityBehavior
+    # @!attribute [r] activity
+    #   @return [Krikri::Activity]
     attr_reader :activity
+
+    ##
+    # @param activity [Krikri::Activity]pp
```
anarchivist: Typo?

no-reply: text editor cruft. Will edit out when doing other revisions.

lib/krikri/entity_behavior.rb
```diff
@@ -21,20 +38,49 @@ def initialize(activity)
...
+    # @see Krikri::EntityBehavior#entities
+    # @see Krikri::Activity#entities
     def self.entities(activity, *args)
       new(activity).entities(*args)
     end
+
+    private
+    
+    ##
+    # Private utility method capturing common logic for applying entity logic to
```
anarchivist: :+1:

lib/krikri/entity_consumer.rb
```diff
@@ -29,20 +35,37 @@ module EntityConsumer
     # because not every activity will modify its entities with provenance
     # messages; an indexing activity, in particular.  In this case an LDP URI
     # representing the activity is not relevant.
-    #
+    #
```
anarchivist: trailing whitespace

lib/krikri/entity_consumer.rb
```diff
@@ -1,15 +1,21 @@
...
-  # such that
-  #   a.generator_activity
-  # returns a Krikri::Activity
+  # A mixin for `Krikri::SoftwareAgent`s that use entities. Provides a 
+  # mechanism for setting an `#entity_source` and consuming entities.
+  #
+  # For backwards compatability, this supports an older interface where entities
+  # are selected based on a `generator_activity`.
+  # 
+  # @example the deprecated interface
```
anarchivist: Do we want to add an `@example` of the new interface?

anarchivist:
> `Mapper::Agent` and `Enricher` are refactored to use the new interface.

Should we do this for `Indexer` as well?

anarchivist:
Looks good so far. I'll assign this to @markbreedlove next.

no-reply:
> Should we do this for Indexer as well?

Yes, but `Indexer` is much more tightly coupled to having direct access to a generator Activity. I don't think there's any specific reason for this, but fixing it will require a near total rewrite of `Indexer`. It seemed better to let this interface shake out, and let an entity source or two emerge before trying to take that on.

Hence the deprecation. :)

anarchivist:
OK, that sounds fine to me, then. Could we then at least add some `@todo` annotations in the Indexer classes to address this in the future?

app/models/krikri/activity.rb
no-reply: As context: `RDF::URI#/` interns the URI.

@anarchivist are you suggesting also doing `RDF::URI.intern(Krikri::Settings['marmotta']['ldp']) / ...`? :+1: to that.

no-reply:
> Could we then at least add some @todo annotations in the Indexer classes to address this in the future?

Can do.

app/models/krikri/activity.rb
anarchivist: Sure. :smile_cat:

lib/krikri/entity_consumer.rb
no-reply: I have an example of the new interface with the first alternative "entity source" in a later commit. Suggest bracketing this for the moment, since that PR is being fixed up now.

no-reply:
This has been fixed up as suggested by @anarchivist and rebased to the latest develop to avoid conflicts.

no-reply:
The failure seems to be Ruby 2.3 specific. Likely to do with the hacked in `include` in the tests. Will try again in the morning.

no-reply:
I'm not sure where the failure discussed in https://github.com/dpla/KriKri/pull/221#issuecomment-190473470 crept in; it doesn't seem to be part of this PR.

Still, it is fixed in 9735f92.

no-reply:
I ended up squashing the first three commits. Thinking to keep 71e482f as a stand alone.

This is in mergeable shape, to my eyes, pending any changes via @markbreedlove's review.  The force pushed versions that are up now are what I'll rebase remaining related work on.

app/models/krikri/activity.rb
```diff
@@ -10,16 +10,41 @@ module Krikri
...
+    ##
+    # @param uri [#to_s] a uri for this activity
+    #
+    # @return [Krikri::Activity] the activity with the given uri
+    #
+    # @raise [RuntimeError] if the URI form does not match the activity
+    # @raise [ActiveRecord::RecordNotFound] if no activity is found
+    def self.from_uri(uri)
+      raise "Cannot find #{self} from URI: #{uri}" unless 
+        uri.start_with? base_uri
```
markbreedlove: Since the condition is that the given URI doesn't start with the base URI, could this message be changed to indicate that the error is with the given URI not matching what we'd like? It hasn't tried to `find` the Activity yet at this point.

markbreedlove:
This all looks good to me. I have that one question about `Activity.from_uri`'s error message, but I'm happy to merge pending a response to that and a rebase onto `develop`.

markbreedlove:
_Bumping back to @no-reply per comment above; bump back to me when it's ready_...

app/models/krikri/activity.rb
no-reply: Is this better?

https://github.com/dpla/KriKri/pull/221/files#diff-e5df2b9df9eba74bdafabb66a4ccfee6R42

no-reply:
I think that should take care of it. Working on getting the next one up, now. :)

markbreedlove:
Sure, that's fine. Thanks!

PR #222, opened 2016-02-25
Feature/travis config

no-reply:
Let's see if this fixes our build problems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,3 +13,4 @@ addons:
   code_climate:
     repo_token: 9b8c58a49bdfba5582d6a766fbe198e3720158e5e65c2ae00b1667e5e1bb8769
 env: CI=yes
+before_install: 'gem install bundler -v "~> 1.11.0"'


.travis.yml
```diff
@@ -13,3 +14,6 @@ addons:
   code_climate:
     repo_token: 9b8c58a49bdfba5582d6a766fbe198e3720158e5e65c2ae00b1667e5e1bb8769
 env: CI=yes
+before_install: 'gem install bundler -v "~> major.minor.patch"'
```
markbreedlove: Is that literally supposed to be "major.minor.patch"?

markbreedlove:
Looks good to me.  We did have an issue with Travis CI failing the 2.3.0 build, only to pass it when it was re-run. Does that cause us any concern?

Otherwise, it would be great to have the bundler version specification added from the other commit.

no-reply:
It seems like the intermittent failure is time dependent.

I'd suggest the way forward is to add the 2.3 build, and dig deeper if/when we see more failures. That the build reveals a potential problem seems like a bad reason to avoid it. :)

markbreedlove:
Sure, I agree. I was just thinking the addition of 2.3, which might be its own feature branch with its own testing phase, might be a separate concern. I could go either way on merging it now or waiting.

no-reply:
I force pushed a change removing the 2.3 build.

PR #223, opened 2016-02-25
Add Ruby 2.3.0 to build list

no-reply:


anarchivist:
Rebased onto `develop`.

no-reply:
:cool:

I'm half anticipating that the timecop/rspec issue we saw last week will resurface, but we can cross that bridge when we come to it. It's clearly a issue with the specs only, in any case.

PR #224, opened 2016-02-26
Feature/entity behavior refactor

no-reply:
Refactors EntityBehavior & SoftwareAgent to use a default entity_behavior at the class level. Sets @entity_behavior on initialization to avoid `||=` in the accessor method.

Since `Krikri::Activity` never alters the `SoftwareAgent`'s entity behavior after initialization, there's no need to initialize it at all. Instead, we simply use the default `EntityBehavior` for the `SoftwareAgent` class.

### DIFF
--- a/app/models/krikri/activity.rb
+++ b/app/models/krikri/activity.rb
@@ -127,14 +127,14 @@ module Krikri
     # that have been affected by this activity.
     #
     # The kind of object that is returned depends on the EntityBehavior class
-    # that is associated with the SoftwareAgent that is represented by the
+    # that is associated with the SoftwareAgent class that is represented by the
     # Activity's `agent' field.
     #
     # @param [Array<Object>] *args Arguments to pass along to
     #                              EntityBehavior#entities
     # @return [Enumerator] Objects
     def entities(*args)
-      agent_instance.entity_behavior.entities(self, *args)
+      agent.constantize.entity_behavior.entities(self, *args)
     end
   end
 end
--- a/lib/krikri/enricher.rb
+++ b/lib/krikri/enricher.rb
@@ -34,14 +34,6 @@ module Krikri
     end
 
     ##
-    # @see Krikri::Activity#entities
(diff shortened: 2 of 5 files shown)


markbreedlove:
OK, that all makes sense.

In answer to your other question, asked elsewhere, I don't know why I had `SoftwareAgent#entity_behavior` set `@entity_behavior ||= nil`. The PR for this was https://github.com/dpla/KriKri/pull/131 and, since I don't see any discussion of it there, I'm not sure what my reasoning was.

This works to remove the requirement that the agent of a generator activity be instantiated.

markbreedlove:
@no-reply I may have merged this too quickly. If I run a mapping where the generator activity's harvester does not exist on the filesystem (in this case, `IaHarvester`), I get `uninitialized constant IaHarvester`. I'll send a followup message by another channel.

no-reply:
@markbreedlove That seems okay to me. I'm not sure how we could get around a dependency on the harvester existing. 

How did you harvest the records initially?

anarchivist:
Can someone confirm #226 fixed the issue described above?

no-reply:
@anarchivist if you mean the `uninitialized constant IaHarvester` issue; #226 is unrelated.

I'm still of the opinion that the `uninitialized constant` error is correct and desirable. I can't see how we can attribute an `Activity` to, and rely on information from, a `SoftwareAgent` class that doesn't exist.

This probably means that if we deprecate a `SoftwareAgent`, we need to tombstone it as a generic agent. Perhaps we can talk about this at check-in tomorrow, or at an upcoming tech team meeting.

anarchivist:
@no-reply Thanks - just wanted to verify that it wasn't the case.

markbreedlove:
@no-reply @anarchivist I agree about it not making sense if the generator's `SoftwareAgent` class doesn't exist.  The issue has become, for me, one of the mappings and the harvesters being kept in separate repositories, where we've found they can become out of sync with one another with our workflow.

PR #225, opened 2016-02-29
Add Smithsonian parser

marktriggs:
This commit got lost along the way, but belongs with the pull request for the Smithsonian mapping: https://github.com/dpla/heidrun-mappings/pull/37

### DIFF
--- a/lib/krikri.rb
+++ b/lib/krikri.rb
@@ -20,5 +20,5 @@ module Krikri
            'krikri/entity_behaviors/aggregation_entity_behavior'
   autoload :OriginalRecordEntityBehavior,
            'krikri/entity_behaviors/original_record_entity_behavior'
-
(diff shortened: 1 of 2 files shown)


anarchivist:
I'm wondering if we actually need this - we might be able to get by with `Krikri::XmlParser` by adding `:parser_args => '//doc'` [here in the mapping](https://github.com/dpla/heidrun-mappings/pull/37/files#diff-1a70b661adb1f79b868dd3204265c0c9R264).

marktriggs:
Yeah, I wasn't sure whether it was desirable just to have the parser as a placeholder or not.  Easy enough to change the mapping though.

anarchivist:
I think this can be closed out once we merge dpla/heidrun-mappings#47.

PR #226, opened 2016-02-29
Fix definition of Krikri::Harvester.entity_behavior

markbreedlove:
Move the definition of `Krikri::Harvester.entity_behavior` to the `included` block, because `Harvester` is a mixin.

Add a test to ensure that a class mixing in Harvester receives `Krikri::OriginalRecordEntityBehavior` from a call to `.entity_behavior`.

### DIFF
--- a/lib/krikri/harvester.rb
+++ b/lib/krikri/harvester.rb
@@ -43,13 +43,13 @@ module Krikri
         def queue_name
           :harvest
         end
-      end
-    end
 
-    ##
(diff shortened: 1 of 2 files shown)


no-reply:
:+1: Will merge if CI is happy.

no-reply:
Noting that Travis showed some unusual Solr problems when running this branch. It looks like a Travis container and/or networking issue, but we should look out for similar errors once this is merged.

PR #227, opened 2016-03-02
ApiHarvester enhancements for CdlApiHarvester

moltude:
Changes to api_harvester.rb to support the CdlApiHarvester implementation in Heidrun. 
* Standardize hash accessors in .next_options()
* Use .deep_dup instead of .dup when calling .request()


lib/krikri/harvesters/api_harvester.rb
```diff
@@ -23,7 +23,7 @@ class ApiHarvester
     # @see .expected_opts
     def initialize(opts = {})
       super
-      @opts = opts.fetch(:api, {})
+      @opts = opts.fetch(:api, {}).symbolize_keys
```
anarchivist: This changes the interface across the other `Krikri::Harvester` implementation classes. By default, I believe the implementation classes expect the outermost "layer" to have symbol-based keys.

lib/krikri/harvesters/api_harvester.rb
```diff
@@ -1,4 +1,4 @@
-module Krikri::Harvesters
+ module Krikri::Harvesters
```
anarchivist: minor: whitespace

anarchivist:
I think we should take a closer look of your implementation of the CDL harvester and see if this is necessary - my guess is that it's not, but we might be running into an issue with what `RestClient` is expecting here.

anarchivist:
> The only change that the CdlApiHarvester requires is the use of .deep_dup when calling .request().

Can you say more about why?

moltude:
In the case of Cdl the parameter as a nested hash of headers and params and .dup does a shallow copy which drops the nested params hash. This could have been overridden in the child class but it seemed like  better anyway.

anarchivist:
Could we look at taking the authorization token value and params as two distinct options for the harvester's instantiation that get reassembled when passed to `RestClient.get`?

moltude:
Absolutely. I was also curious why .deep_dup was used here (https://github.com/dpla/KriKri/blob/develop/lib/krikri/harvesters/api_harvester.rb#L138) but not on 141. Can we drive into that as well?

PR #228, opened 2016-03-04
Raise error for empty original record, minting ID

markbreedlove:
Raise an exception when minting an ID in `DPLA::MAP::Aggregation#mint_id!` if there is no original record, instead of generating a random one.

As previously written, if a mapping failed to map `originalRecord`, you would end up with two `RDFSource`s, one for the harvested record, and one for the mapped record with a random ID.

Also improve the existing error messaging for `DPLA::MAP::Aggregation#original_record` when there's no original record.

### DIFF
--- a/lib/krikri/engine.rb
+++ b/lib/krikri/engine.rb
@@ -129,13 +129,13 @@ module Krikri
         # Get the persisted original record for this Aggregation.
         # @return [Krikri::OriginalRecord, nil]
         #
-        # @raise [NameError] when the original record id is not a URI or the
-        #   the original record has not been persisted to Marmotta.
+        # @raise [NameError] when the original record is empty or is a blank
+        #   node.
         #
         # @raise [Faraday::ConnectionError] when there is a connection problem
         #   with Marmotta.
         def original_record
-          raise NameError, "#{originalRecord.first} is not an OriginalRecord" if
+          raise NameError, no_origrec_message if
             originalRecord.empty? || originalRecord.first.node?
           Krikri::OriginalRecord.load(originalRecord.first.rdf_subject
             .to_s)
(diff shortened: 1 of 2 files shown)


markbreedlove:
I'd recommend waiting until https://github.com/dpla/KriKri/pull/229 is merged before merging this.

no-reply:
Looks okay.

Would it be a good idea to replace `NameError`, like we did in https://github.com/dpla/KriKri/commit/3b36c71e8d898fff206d965c64b62ef20291e5a5#diff-fbfeb4a886bf59e266597943d1f9baedR208? `#no_origrec_message` feels like a Single Responsibility violation. Or perhaps these just need to be private methods?

markbreedlove:
`no_origrec_message` already is a private method.  I'm finding that it's going to be a little time-consuming to figure out a good way to rework it to use a custom exception, because of the way `DPLA::MAP::Aggregation` is monkeypatched; though I agree that `NameError` is not ideal.

lib/krikri/engine.rb
```diff
@@ -164,10 +164,24 @@ def local_name_from_original_record
...
+          # if a mapping failed to map it.
+          raise NameError, no_origrec_message if
             originalRecord.empty? || originalRecord.first.node?
           local_name_from_original_record
         end
+
+        def no_origrec_message
+          "#{dpla_id} #{no_origrec_cond}"
+        end
+        def no_origrec_cond
```
no-reply: missing empty line between methods

no-reply:
> I'm finding that it's going to be a little time-consuming to figure out a good way to rework it to use a custom exception, because of the way DPLA::MAP::Aggregation is monkeypatched; though I agree that NameError is not ideal.

There is an example right above at: https://github.com/dpla/KriKri/blob/fix/id-minting-no-origrec/lib/krikri/engine.rb#L94-L98

It should also be possible to nest an error class into the monkey patched `Aggregation`. I do agree that this really needs to be moved out of an initializer, though.

I'm also okay pushing this through as is, pending the minor formatting fixup at https://github.com/dpla/KriKri/pull/228#discussion_r55541295

markbreedlove:
I've fixed the whitespace and rebased.

My issue with the exception class was that I was running into trouble putting it in right module. I didn't want it to be in the root namespace; I think I was trying to put it in `DPLA::MAP` because it pertains to `DPLA::MAP::Aggregation`, and I was getting undefined constant errors. I think it's something I need to refine another day.

no-reply:
#231 addresses the test that failed in a previous run.

PR #229, opened 2016-03-04
Fix harvester_spec.rb test with wrong expect subject

markbreedlove:
In Harvester spec, fix ".entity_behavior knows its entity_behavior" to expect `subject.entity_behavior` instead of `klass.entity_behavior`.

That should normally fail, because `klass` should not be expected to have `include`d `Krikri::Harvester` yet.  Depending on the order in which the specs were run, the issue was being masked, because the spec was operating on a `klass` that had had `Harvester` included by an earlier spec.

### DIFF
--- a/spec/lib/krikri/harvester_spec.rb
+++ b/spec/lib/krikri/harvester_spec.rb
@@ -36,7 +36,8 @@ describe Krikri::Harvester do
 
   describe '.entity_behavior' do
     it 'knows its entity behavior' do
-      expect(klass.entity_behavior).to eq(Krikri::OriginalRecordEntityBehavior)
+      expect(subject.class.entity_behavior)
+        .to eq(Krikri::OriginalRecordEntityBehavior)
     end
   end
 


anarchivist:
:+1:

no-reply:
Nice work spotting the problem, here.

I think this test really does mean to be testing the class method, though. The description uses the `.` method prefix (`describe '.entity_behavior'`), and the class method is what will be called by the Activity.

Maybe we should do `expect(subject.entity_behavior).to eq(Krikri::OriginalRecordEntityBehavior)`.

markbreedlove:
> Maybe we should do `expect(subject.entity_behavior).to eq(Krikri::OriginalRecordEntityBehavior)``.

That's actually what the change is doing: https://github.com/dpla/KriKri/blob/fix/harvester-spec-wrong-subject/spec/lib/krikri/harvester_spec.rb#L36

Since the test is really testing `Krikri::Harvester` to make sure that the `entity_behavior` that's defined in that class returns the right thing, you have to use `subject` in the spec like the other examples in the spec.

no-reply:
Sorry, my suggestion should have read `subject.class.entity_behavior`. The current test checks the instance method, but the test intends to check the class method.

no-reply:
Looks good. :squash: :question:

PR #230, opened 2016-03-08
Feature/count query

no-reply:
Adds support for `ProvenanceQueryClient#count_by_activity` using
SPARQL's `COUNT` functionality. Refactors `ProvenanceQueryClient`
significantly in the process.

`Krikri::Activity` gets an `#entity_count` method, which is much faster
in practice than `#entity_uris.count`. This allows optimized counting on
the server side, instead of returning an entire result set and iterating
through the `RDF::Query::Solutions`.

Note: this commit is WIP. The pending test on `#count_by_activity` is
critical, but fails on H2 when the `|` operator is used in the
predicate. This is being pushed up to track the WIP, but needs testing
in a more production like environment before it can be merged. It may be
that we'll need to report a bug on Marmotta and leave the test marked as
pending, for the moment.

__I intend to deploy this to staging after check-in tomorrow (3/9) to test the 
pending example with the real database backend. The generated query 
works against the production system, but not in testing; I want to isolate H2
as a possible cause.__

### DIFF
--- a/app/models/krikri/activity.rb
+++ b/app/models/krikri/activity.rb
@@ -102,6 +102,13 @@ module Krikri
     end
 
     ##
+    # @return [Integer] count of entities changed by this activity
+    def entity_count(include_invalidated = false)
+      Krikri::ProvenanceQueryClient
+        .count_by_activity(self.to_term, include_invalidated)
+    end
+
+    ##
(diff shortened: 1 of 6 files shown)


no-reply:
Closing this pending a fix for the failing test.

PR #231, opened 2016-03-10
Remove Solr docs more carefully in test suite

no-reply:
The particular issue was in the specs for
`Krikri::QASearchIndex#update_from_activity`. These got left behind,
some tests had apparently compensated for this by cleaning up in
`before`. That behavior is removed, to make it easier to catch issues
like this in the future.

Also normalizes solr index cleanup on the `clear_search_index` helper. I
think this was added after some of these specs were written, and some of
them never got migrated.

### DIFF
--- a/spec/lib/krikri/indexer_spec.rb
+++ b/spec/lib/krikri/indexer_spec.rb
@@ -13,7 +13,9 @@ describe Krikri::Indexer do
   behaves_opts = { generator_uri: enrichment_gen_uri,
                    index_class:   'Krikri::QASearchIndex' }
 
-  it_behaves_like 'a software agent', behaves_opts
+  it_behaves_like 'a software agent', behaves_opts do
+    after { clear_search_index }
+  end
   
   # See mapper_agent_spec.rb regarding :opts and behaves_opts...
   let(:opts) do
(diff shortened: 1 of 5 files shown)


no-reply:
I was able to track this down by using the seed from the failed travis run: `bundle exec rspec --seed 52769`

no-reply:
The second test fixes what seems to be a new issue with webmock/CodeClimate. CodeClimate messages were going out before the test suite ran (but only sometimes?), and webmock blocked them.

markbreedlove:
That looks good to me!  I'll merge it, if nobody else wants to look.  I'll just bump it over to @anarchivist in case he wants to comment or pass it around.

spec/lib/krikri/indexer_spec.rb
```diff
@@ -13,7 +13,9 @@
   behaves_opts = { generator_uri: enrichment_gen_uri,
                    index_class:   'Krikri::QASearchIndex' }
 
-  it_behaves_like 'a software agent', behaves_opts
+  it_behaves_like 'a software agent', behaves_opts do
+    after { clear_search_index}
```
anarchivist: minor: add a space before `}`

no-reply: :+1:

anarchivist:
I think this looks good other than the minor change I identified.

no-reply:
Fixed and squashed back into 8198c90.

PR #232, opened 2016-03-14
Handle EDTF Intervals

no-reply:
Intervals do not have their own precision, but instead rely on the precision of their begin and end date/time. This fixes a `NoMethodError` on `EDTF::Interval#uncertain!` and adds coarse handling for interval uncertainty.


moltude:
looks good

PR #233, opened 2016-03-17
Add support for "inline exceptions"

marktriggs:
If this option is set to true, connection-level errors (like
CONNREFUSED, socket timeouts, etc.) will be delivered to the
caller as HTTP 5xx responses.

-----
The context to this is https://github.com/dpla/heidrun/pull/38, where for the IA harvester we want to ignore connection refused/timeout exceptions and just skip over the affected records.

The IA harvester already skips records when it gets back a 5xx error response, so we really just want it to do the same thing for exceptions caused by connection errors.  It seemed a shame to complicate the harvester code with a bunch of exception handling--especially when future harvesters will likely want the same thing--so I added the option for "inline exceptions" to give the caller the option of handling all types of errors in a uniform way.

The change is backward compatible, so existing harvests shouldn't be affected.  If this looks OK I'll update the IA harvester to use the new option, and then all should be good in the world.

### DIFF
--- a/lib/krikri/async_uri_getter.rb
+++ b/lib/krikri/async_uri_getter.rb
@@ -51,6 +51,8 @@ module Krikri
     #   redirects.
     # @option opts [Integer] :max_redirects Number of redirects to follow before
     #   giving up. (default: 10)
+    # @option opts [Boolean] :inline_exceptions If true, pass exceptions as a
+    #   5xx response with the exception string in the body. (default: false)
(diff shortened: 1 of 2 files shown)


markbreedlove:
This looks good to me. I know that we often try to have one expectation per test, but I'll let others weigh in on that.  I'll pass this to the next reviewer. Thanks!

anarchivist:
I think this looks good - thanks!

PR #234, opened 2016-03-17
Only inline exceptions stemming from Faraday

marktriggs:
Previously an exception thrown in the body of a `with_request` block
would be caught and propagated as an inline exception.  This can be
quite confusing when your `with_request` block is suddenly being called
twice!

Added a test to catch this case too.

-----
Only too late do I realize I should have called this branch "inline-exception-inception".  Damn.


anarchivist:
:+1:

PR #235, opened 2016-04-08
Add SplitOnProvidedLabel enrichment

no-reply:
Splits a resource into multiple resources when multiple providedLabel are present.

This is useful when a mapping error or limitation in partner data makes it impossible/impractical to split resources properly at mapping time. This is the case with NARA's subject mapping (it's not clear yet whether "impossible/impractical" is the case--this may simply be our mistake).

### DIFF
--- a/lib/krikri/enrichments/split_on_provided_label.rb
+++ b/lib/krikri/enrichments/split_on_provided_label.rb
@@ -1,0 +1,58 @@
+module Krikri::Enrichments
+  ##
+  # Splits a resource into multiple resources when multiple providedLabels are
+  # present. This is useful when a mapping error or limitation in partner data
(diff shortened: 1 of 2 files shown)


lib/krikri/enrichments/split_on_provided_label.rb
```diff
@@ -0,0 +1,58 @@
+module Krikri::Enrichments
+  ##
+  # Splits a resource into multiple resources when multiple providedLabels are
+  # present. This is useful when a mapping error or limitation in partner data
+  # makes it impossible to split resoureces properly at mapping time.
```
anarchivist: minor: typo

anarchivist:
:tada:

PR #236, opened 2016-04-11
Fix/description map crosswalk

no-reply:


no-reply:
Going to let this pass before squashing it, just in case. :)

moltude:
:+1:

PR #237, opened 2016-04-11
Make language code matching case insensitive

no-reply:
`LanguageToLexvo` now matches three-letter language codes independent of case.


moltude:
:+1:

PR #239, opened 2016-05-27
Remove guard dependency

no-reply:
Guard has stopped supporting Ruby 2.1. I'm removing the dependency, pending an upgrade on our end.

See:
https://github.com/guard/guard#important-ruby-21-is-officially-outdated-and-unsupported-please-upgrade-to-ruby-23-or-225-before-installing-guard


AudreyAltman:
:+1:  Merge, or do you want others to look at this as well?

no-reply:
Seems like a :merge: to me.

PR #240, opened 2016-06-07
Add array index access; mimics `Array#at`

no-reply:
Allows index and range access on `Krikri::Parser::ValueArray` objects, maintaining method chaining.

### DIFF
--- a/lib/krikri/parser.rb
+++ b/lib/krikri/parser.rb
@@ -213,6 +213,14 @@ module Krikri
       end
 
       ##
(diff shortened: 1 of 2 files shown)


no-reply:
This is in part to address the need raised at https://github.com/dpla/heidrun-mappings/pull/54. That PR should be able to move forward with the direct `#[]` call discussed there, but it would be good to flag changing to `#at` as a To Do item.

moltude:
👍

PR #241, opened 2016-06-08
Add `#compact` to mapping DSL

no-reply:
Makes it possible to squash out `nil` values in value array with `#compact`.


moltude:
👍

PR #242, opened 2016-06-13
Migrate to using krikri-spec

no-reply:
Moves common shared examples and matchers out to a separate gem. This
allows reuse of the test infrastructure in applications using the KriKri
engine.

After this change, some development workflows on KriKri will change:

  - when writing tests or test helper code, consider whether the work
  will be useful to implementing apps. If so, move it to `krikri-spec`.
  - `krikri-spec` should be kept up-to-date. to this end, it is pinned
  to the `develop` branch on GitHub in the `Gemfile`.
  - It will sometimes be useful to use local copies of `krikri-spec`,
  especially when working on tests as described above. To do this, it's
  recommended that you use Bundler's `path:` option. See:
  https://github.com/dpla/krikri-spec#development


moltude:
Looks good, some tests are failing but it matches the behavior of the old integrated tests so I think this is 👍  and the failed tests will be worked out separately.

no-reply:
@moltude which tests are failing? 

CI seems to be passing.

moltude:
I should have clarified that -- I'm pretty sure my failures are a result of local issues since I get the same failures for `krikri-spec` and the old integrated tests. I'm failing a slew of tests that appear to be related to some a mis-configuration within Solr resulting in 400 errors. The tests are complaining about 'type' not being a known field within Solr.

moltude:
`bundle exec rake  jetty:config` resolved 95% of those errors. Now the only three tests that fail are 

```
rspec ./spec/lib/krikri/mapping_dsl/parser_methods_spec.rb:51 # Krikri::MappingDSL::ParserMethods#record_uri raises an error when record does not exist
rspec ./spec/lib/krikri/mapping_dsl/rdf_subjects_spec.rb:122 # Krikri::MappingDSL::RdfSubjects Krikri::MappingDSL::RdfSubjects::SubjectDeclaration#to_proc with too many values raises an error
rspec ./spec/models/dpla/map/aggregation_spec.rb:78 # DPLA::MAP::Aggregation#mint_uri! with originalRecord with more than one originalRecord raises an error
```

PR #244, opened 2016-06-21
Feature/dsl backtrack

no-reply:
Allows binding arbitrary points in the tree to variables for later
recall. Initially supports an `#and(from: :var)` construction which
recalls the variable if the current node is empty.

This is similar to `#if`/`#else` but with more flexibility.

----

Refactors existing `#if`/`#else` and other `:top` calls to use the more general underlying interface.

----

Note that the underlying `bindings` hash is shared between the many `ValueArray` nodes in a given call chain, so patterns like `record.field(:field).bind(:subject).field(:nested).bind(:subject_label).reject { |_| true }.and(from: :subject)` will leave `:subject_label` bound in the result. I.e. further conditions could result in returning to the label.

----

Closes https://issues.dp.la/issues/8184

### DIFF
--- a/lib/krikri/parser.rb
+++ b/lib/krikri/parser.rb
@@ -157,13 +157,14 @@ module Krikri
     # of literal values (retrieved from Parser::Value#value), or a single
     # literal value.
     #
-    # Uses `@top` to track a base node for recovery via `#else`. Methods that 
-    # return a `ValueArray` should pass `@top` down to the new instance.
+    # Uses `@bindings` to track variables for recovery via `#else`, `#from`, and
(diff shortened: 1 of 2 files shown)


moltude:
@no-reply I think I understand most of this and it looks 👍 to me but what exactly is deprecated on line 194?

no-reply:
> what exactly is deprecated on line 194?

The second argument to the initializer has changed. You used to simply pass another `ValueArray` which would be set to `@top`; now you should pass a set of bindings which may include a `:top`. The old way is still supported (per SemVer) through the conditional.

moltude:
👍  to merge

PR #245, opened 2016-06-29
Refactor & Document MappingDSL #to_procs

no-reply:
MappingDSL Declarations were somewhat messy and under-documented, particularly the `#to_proc` methods. This refactors slightly, removing some undue complexity. It also documents the Declaration classes much more thoroughly.

The second commit satisfies rubocop about the related integration tests in `mapper_spec.rb`.


moltude:
@no-reply thanks for adding all that documentation, it has made these methods much clearer and is really helpful. 👍 to merge

PR #246, opened 2016-06-30
Force Nokogiri security upgrade

no-reply:
Nokogiri had security vulnerabilities in versions `>= 1.6.0`, `< 1.6.8`. Since Blacklight requires `~> 1.6`, it seems good to force engine users to use `1.6.8` or greater.

Our production instance upgraded in a deployment some time ago.


markbreedlove:
Looks good!

PR #247, opened 2016-07-05
Map providedLabel in TimeSpan

no-reply:
TimeSpan mappings in the DPLA MAP v3.1 crosswalk need to map labels aggressively. If no label is given, the frontend won't display a date. This maps `prefLabel` with preference, but uses `providedLabel` as a backup.

Closes https://issues.dp.la/issues/8457.


AudreyAltman:
:rocket:

PR #248, opened 2016-07-08
Add TimespanLabel enrichment

no-reply:
The new enrichment sets a `prefLabel` based on existing begin/end dates
within a TimeSpan.

Does nothing if:

  - a `prefLabel` is already present
  - neither a begin, nor end date is given

### DIFF
--- a/lib/krikri/enrichments/timespan_label.rb
+++ b/lib/krikri/enrichments/timespan_label.rb
@@ -1,0 +1,48 @@
+module Krikri::Enrichments
+  ##
+  # Builds and sets a `prefLabel` based on existing begin/end dates within an 
+  # `edm:TimeSpan`.
+  #
+  # @example
+  #   date = DPLA::MAP::TimeSpan.new.tap do |t| 
+  #     t.begin = Date.today
+  #     t.end = (Date.today + 1)
+  #   end
+  #
+  #   TimespanLabel.new.enrich_value(date).prefLabel
+  #   # => ["2016-07-08/2016-07-09"]
+  #
+  class TimespanLabel
+    include Audumbla::FieldEnrichment
+
+    ##
+    # Add a prefLabel for `DPLA::MAP::TimeSpan` objects with begin/end dates
+    #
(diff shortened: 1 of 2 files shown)


no-reply:
@guegueng the labels generated look like:

  - `2016-07-08` if begin/end dates are the same, or if only one is present
  - `2016-07-01/2016-07-31`if it is an interval.

We could probably refine cases like the latter to simply `2016-07`, for full months, or `2016` for spans that cover a full year. Does the current format work okay?

guegueng:
yes, the formatting is fine with me

moltude:
Okay, I'll dig in right after the staff meeting and hope to get this close out before lunch

moltude:
I think this looks good for a release. Will start by running against production WI, DigComm, ESDN (anyone else in need of this enrichment)?

guegueng:
you can add it to PA and to the standard enrichment profile

guegueng:
actually, I think you can skip PA, they only use a single field for date. Add it to the standard enrichment profile though. Most Hubs aren't restrictive like PA is.

no-reply:
Note that where this gets added to the profile matters. We have existing enrichments that copy `providedLabel` to `prefLabel`, and this will only generate a `prefLabel` if one does not already exist.

There's a policy question here: For the standard profile, if we have mapped both a `providedLabel` and `begin/end` dates, do we want to prefer the given `providedLabel` or the generated (standardized) label?

guegueng:
I the generated one in this case

moltude:
@guegueng can you clarify the last comment?

guegueng:
I think we should use the prefLabel, or the generated standard one for the display.

moltude:
@guegueng Just so I am sure I understand this correctly. For the default/standard profile, we run the `TimespanLabel` enrichment first to generate a standardized label `prefLabel` and then run the `CreatePrefLabelFromProvided ` enrichment to copy providedLabel to prefLabel only for those records where a standardized `prefLabel` could not be created. Do I have that right?

PR #249, opened 2016-07-13
Fixup build_place 

moltude:
* Map providedLabel first and then label only if no providedLabel exists. 
* Fixup the spec test for map_crosswalk
* Misc. whitespace cleanup


no-reply:
:100:

PR #250, opened 2016-07-13
Clean up dependencies

no-reply:
Some of the dependencies were sloppily specified, this cleans them up to
be a bit more precise. Some versions are given, where none were
specified before. Others are loosened up to depend on a major version,
but with a minimum version specified (e.g. from '~>0.0.6' to '~>0.0',
'>=0.0.6').

Blacklight is left alone, since I'm pretty sure we can't trust minor
versions to be free upgrades.


no-reply:
Closing this until I have a chance to figure out what is going on, here.

PR #251, opened 2016-07-18
Upgrade EDTF to 3.0

no-reply:
Upgrades the `edtf` gem and locks to the most recent major version.

The new version was released over the weekend.


anarchivist:
Needs a rebase.

no-reply:
Rebased.

PR #252, opened 2016-07-18
Mapping DSL match/reject based on child node.

no-reply:
Introduces a syntax for matching or rejecting nodes based on the
presence of a child node.

For instance:

    record.field('mods:subject').reject_child('mods:hierarchicalGeographic')

Closes https://issues.dp.la/issues/8207


moltude:
👍

PR #253, opened 2016-07-18
Fix RSpec warning on `raise_error`

no-reply:
The `raise_error` matcher wants to check against a real error message to
avoid false positives. In addition to giving it a parameter, I needed to
fix a false positive (via WebMock) to make this test pass.


moltude:
👍

PR #254, opened 2016-07-21
Use providedLabel in TimeSpan for MAP 3.1

no-reply:
Use providedLabel with preference, falling back on prefLabel if needed.


moltude:
👍

PR #255, opened 2016-08-01
facets for pref and provided labels

AudreyAltman:
For all existing facets in the QA interface that have both `prefLabel` and `providedLabel`, this makes facets for both.  One exception is `language`, which has a `providedLabel` but no `prefLabel` in the solr schema.

For most fields, `name` in the solr schema corresponds with `prefLabel`, so `sourceResource_type_name` is the `prefLabel` for `type`.  The one exception seems to be `collection`, for which `title` corresponds to `prefLabel`.

There are no tests b/c this uses Blacklight functionality.  There are currently no tests in `Krikri` for faceting behavior.  This has been tested locally using sample data.  It addresses [ticket #8501](https://issues.dp.la/issues/8501).

### DIFF
--- a/app/controllers/krikri/records_controller.rb
+++ b/app/controllers/krikri/records_controller.rb
@@ -52,28 +52,46 @@ module Krikri
       # facet bar
 
       config.add_facet_field 'sourceResource_type_name',
-                             label: 'Type',
+                             label: 'Type (pref)',
+                             limit: 20
+      config.add_facet_field 'sourceResource_type_providedLabel',
+                             label: 'Type (provided)',
                              limit: 20
       config.add_facet_field 'sourceResource_format',
                              label: 'Format',
                              limit: 20
       config.add_facet_field 'sourceResource_language_providedLabel',
-                             label: 'Language',
+                             label: 'Language (provided)',
+                             limit: 20
+      config.add_facet_field 'sourceResource_spatial_name',
+                             label: 'Place (pref)',
                              limit: 20
       config.add_facet_field 'sourceResource_spatial_providedLabel',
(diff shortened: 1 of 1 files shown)


no-reply:
This looks good.

Do you have any ideas about testing... Some of the changes required to update `DPLA::MAP` to ActiveTriples `0.10+` will change some of these facet paths (`*_name`). I'm pretty sure missing that step would leave us a broken code base that passes CI. If we can come up with a plan for testing, I could get it in as part of that upgrade.

AudreyAltman:
@no-reply I think in that case we would want to write a test for the [sorl schema](https://github.com/dpla/KriKri/blob/develop/solr_conf/schema.xml).  The facets will throw an error if the facet field not exist in the schema, but it makes no difference whether or not that facet field is actually represented in the data.

If the facet field doesn't match the schema, it throws a `RSolr::Error::Http - 400 Bad Request Error`, so I can write a test checking for that - but I don't know that this will address your concern about changes to `DPLA::MAP`.

AudreyAltman:
@no-reply upon further inspection I found that several controller tests will already fail if the facet field doesn't match the schema, so I think we're covered as far as we can be.

no-reply:
@AudreyAltman :ok_hand: i think I can probably whip up a linter for the schema that checks the fields actually resolve to something. It looks like we're already doing most of the work in `schema_keys` https://github.com/dpla/KriKri/blob/develop/lib/krikri/search_index.rb#L226-L236.

> @no-reply upon further inspection I found that several controller tests will already fail if the facet field doesn't match the schema, so I think we're covered as far as we can be.

Then I say :shipit:

PR #256, opened 2016-08-03
Restores `ValueArray` in scoped DSL variables

no-reply:
"Variables" declared as scoped DSL methods were mistakenly cast to their
`#value` when we want a `ValueArray`. This restores the old behavior and
adds an integration test.

This could still use a significant refactor. There is probably a way to
do this without nested iterations through the values. Additionally,
`RecordProxy#dup` is customized in a way that makes this somewhat
tightly coupled and hard to read.

Fixes https://issues.dp.la/issues/8522.

### DIFF
--- a/lib/krikri/mapping_dsl/child_declaration.rb
+++ b/lib/krikri/mapping_dsl/child_declaration.rb
@@ -84,11 +84,15 @@ module Krikri::MappingDSL
         if each_val
           iter = each_val.respond_to?(:call) ? each_val.call(record) : each_val
           iter.each do |value|
-            map = ::Krikri::Mapping.new(target_class)
+            map = Krikri::Mapping.new(target_class)
 
-            # define as_sym on only this instance
+            # define as_sym to return the node (not the value) for this value, 
+            # only on this instance
             map.define_singleton_method(as_sym) do
-              value.respond_to?(:value) ? value.value : value
+              each_val.dup.select do |v|
+                v = v.value if v.respond_to? :value
+                v == value
+              end
             end
 
             map.instance_eval(&block)
@@ -97,7 +101,7 @@ module Krikri::MappingDSL
         # else, process a single child mapping over a single instance of 
         # `target_class`
         else
-          map = ::Krikri::Mapping.new(target_class)
+          map = Krikri::Mapping.new(target_class)
           map.instance_eval(&block)
           target.send(setter, map.process_record(record))
         end
--- a/spec/lib/krikri/mapper_spec.rb
+++ b/spec/lib/krikri/mapper_spec.rb
@@ -13,6 +13,12 @@ describe Krikri::Mapper do
             providedLabel record.field('dc:creator')
           end
 
+          rightsHolder :class => DPLA::MAP::Agent,
+                       :each => record.bind(:rec).field('dc:creator').and(from: :rec),
+                       :as => :ident do
+            providedLabel ident.field('dc:creator')
+          end
+
           contributor :class => DPLA::MAP::Agent,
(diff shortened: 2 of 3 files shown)


markbreedlove:
This is close, but I have a mapping that I still can't get to work, as follows.

The original record (excerpt):
```
<name type="personal">
    <namePart>Garcke, Aug. (August),</namePart>
    <namePart type="date">1819-1904</namePart>
    <role>
        <roleTerm type="text">test</roleTerm>
    </role>
</name>
<name type="personal">
    <namePart>
        Schlechtendal, D. F. L. von (Diedrich Franz Leonhard von),
    </namePart>
    <namePart type="date">1794-1866</namePart>
    <role>
        <roleTerm type="text">contributor</roleTerm>
    </role>
</name>
<name type="personal">
    <namePart>
        Schlechtendal, D. F. L. von (Diedrich Franz Leonhard von),
    </namePart>
    <namePart type="date">1794-1866</namePart>
    <role>
        <roleTerm type="text">test</roleTerm>
    </role>
</name>
```
... Where only one contributor should be selected, the one with "contributor" in `roleTerm`.

My mapping:
```
    contributor class: DPLA::MAP::Agent,
                each: record.field('mods:name')
                            .bind(:name)
                            .field('mods:role', 'mods:roleTerm')
                            .select { |v|
                              v.value.downcase == 'contributor'
                            }
                            .and(from: :name),
                as: :contrib do
      providedLabel contrib.field('mods:namePart')
                           .reject_attribute(:type, 'date')
    end
```

Result:
```
    <http://www.europeana.eu/schemas/edm/aggregatedCHO> [
      a <http://dp.la/about/map/SourceResource>;
      <http://purl.org/dc/terms/contributor> [
        a <http://www.europeana.eu/schemas/edm/Agent>;
        <http://dp.la/about/map/providedLabel> "Garcke, Aug.   (August),"
      ],  [
        a <http://www.europeana.eu/schemas/edm/Agent>;
        <http://dp.la/about/map/providedLabel> "Schlechtendal, D. F. L. von   (Diedrich Franz Leonhard von),"
      ],  [
        a <http://www.europeana.eu/schemas/edm/Agent>;
        <http://dp.la/about/map/providedLabel> "Schlechtendal, D. F. L. von   (Diedrich Franz Leonhard von),"
      ]
    ];
```

The mapping succeeds in taking the `namePart` text nodes where attribute `type` is not "date", but fails to select only the one that has `roleTerm` of "contributor".

It's not clear to me how the `select` could return `true` for all three elements.

no-reply:
I can try to look at that mapping later today. Are you thinking the mapping issue is connected to https://issues.dp.la/issues/8522?

markbreedlove:
Right, I thought that was the inspiration for this PR.

no-reply:
Yes. I just wanted to check on whether the above mapping looks like a separate issue or not. I'm thinking it is, but happy to hold this PR and that 8522 until we can conform/resolve it.

markbreedlove:
I wasn't sure what to make of it, exactly, and haven't been doing mappings a lot lately, so wasn't sure when the behavior changed. I was assuming it had something to do with "rewinding" to the `:name` node. I can confirm that I'm at least not getting any more errors thrown with the mapping code above.

AudreyAltman:
This looks okay to me, but we should get @moltude's opinion as well, since he is more familiar with the larger context of this change.

no-reply:
Rebased.

There was a bit of confusion about the test suite, related to Solr schema changes. Local branches trying to build after #258 will need to get a fresh engine cart app. An easy way to do this is `rake ci`.

moltude:
👍 and merging

PR #257, opened 2016-08-03
Avoid calling `Registry#get` repeatedly

no-reply:
There's no need to lookup the mapping in the registry for each record. Just look it up once and hold the mapping in memory.

This is a minor optimization, but it was easy and might make a difference in GC.


AudreyAltman:
👍 rebase?

no-reply:
Rebased.

PR #259, opened 2016-08-10
add more field value reports

AudreyAltman:
This adds more fields for generating field value reports.
It also corrects the key name for `sourceResource_alternative`.

This has been tested locally.  It addresses [ticket #8205](https://issues.dp.la/issues/8502).

### DIFF
--- a/spec/models/field_value_report_spec.rb
+++ b/spec/models/field_value_report_spec.rb
@@ -10,7 +10,7 @@ describe Krikri::FieldValueReport do
 
   let(:agg) do
     p = DPLA::MAP::Agent.new(RDF::URI(provider_base) / provider_id)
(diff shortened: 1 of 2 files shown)


AudreyAltman:
All tests passing.

spec/models/field_value_report_spec.rb
```diff
@@ -68,8 +68,7 @@
     end
 
     it 'returns "__MISSING__" for field without value' do
-      r = described_class.find('sourceResource_alternative_providedLabel',
-                               provider_id)
+      r = described_class.find('sourceResource_date_name', provider_id)
```
no-reply: Another case where we need a `Maybe` monad.

image

no-reply:
:+1:

PR #263, opened 2016-08-18
add gallery view for search results

AudreyAltman:
This adds a gallery view for search results, using the blacklight-gallery gem.

Since we are still using blacklight 5, the most recent version of blacklight-gallery we can use is 0.3.0.  Unfortunately, blacklight-gallery 0.3.0 fails with ruby 2.3.0, so we have to allow failure for ruby 2.3.0 until we can upgrade our dependencies.  [Ticket 8534](https://issues.dp.la/issues/8534) addresses this issue.

As requested by GG, the gallery view only shows the image - no additional metadata is displayed.

This has been tested locally.  It addresses [ticket #8504](https://issues.dp.la/issues/8504).

### DIFF
--- a/app/controllers/krikri/records_controller.rb
+++ b/app/controllers/krikri/records_controller.rb
@@ -140,6 +140,8 @@ module Krikri
       config.show.route = { controller: 'records' }
 
       config.solr_document_model = Krikri::SearchIndexDocument
+
+      config.view.gallery.partials = []
     end
 
     private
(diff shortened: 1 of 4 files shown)


no-reply:
Let's submit the Ruby 2.3 drop as a separate PR (or at least a separate commit), so that decision is clear and we can revert it easily later.

AudreyAltman:
Allowing failure for ruby 2.3.0 has been dropped from this PR and submitted in a separate PR here: https://github.com/dpla/KriKri/pull/264

AudreyAltman:
Rebased to develop.

PR #264, opened 2016-08-18
allow failure for ruby 2.3.0

AudreyAltman:
Allow failure for ruby 2.3.0.  This is a prerequisite for adding the blacklight-gallery gem with our current dependencies.  See https://github.com/dpla/KriKri/pull/263


no-reply:
This represents a drop of support for Ruby 2.3.0. It also makes a Blacklight upgrade a requirement for https://issues.dp.la/issues/8302.

no-reply:
:+1:

no-reply:
I updated https://issues.dp.la/issues/8302 with Blacklight 6 as a blocker.

no-reply:
Away we go! I think this makes 8302 and it's new blocker much more pressing, as this represents very real tech debt.

PR #265, opened 2016-08-25
pin gems to minor versions

AudreyAltman:
Pin gems to tiny versions currently running in heidrun production.

This addresses (in part) [ticket #8538](https://issues.dp.la/issues/8538).

### DIFF
--- a/krikri.gemspec
+++ b/krikri.gemspec
@@ -23,25 +23,25 @@ Gem::Specification.new do |s|
   s.test_files = Dir["spec/**/*"]
 
   s.add_dependency "rails", "~> 4.1.6"
-  s.add_dependency "rails_config", "~>0.4.0"
-  s.add_dependency "audumbla", '~> 0.1'
+  s.add_dependency "rails_config", "~> 0.4.0"
+  s.add_dependency "audumbla", "~> 0.2"
   s.add_dependency "rdf-turtle", "~>1.1.8"
   s.add_dependency "rdf", "~> 1.1.13"
   s.add_dependency "dpla-map", "4.0.0.0.pre.13"
-  s.add_dependency "rest-client"
-  s.add_dependency "rdf-marmotta", '>= 0.0.6'
-  s.add_dependency "blacklight", "~>5.8.0"
-  s.add_dependency "therubyracer"
+  s.add_dependency "rest-client", "~> 2.0"
+  s.add_dependency "rdf-marmotta", "~> 0.0.6"
+  s.add_dependency "blacklight", "~> 5.8.0"
+  s.add_dependency "therubyracer", "~> 0.12"
   s.add_dependency "edtf", '~> 3.0'
-  s.add_dependency "text"
-  s.add_dependency "oai", '~>0.4.0'
-  s.add_dependency "jsonpath"
-  s.add_dependency "devise", "~>3.4.1"
-  s.add_dependency "resque", "~>1.0"
+  s.add_dependency "text", "~> 1.3"
+  s.add_dependency "oai", '~> 0.4.0'
+  s.add_dependency "jsonpath", '~> 0.5'
+  s.add_dependency "devise", "~> 3.4.1"
+  s.add_dependency "resque", "~> 1.26"
   s.add_dependency "analysand", "4.0.0"
-  s.add_dependency "yajl-ruby"
-  s.add_dependency "elasticsearch", "~>0.4.0"
-  s.add_dependency "nokogiri", ">=1.6.8"
+  s.add_dependency "yajl-ruby", "1.2"
+  s.add_dependency "elasticsearch", "~> 0.4.0"
+  s.add_dependency "nokogiri", "~> 1.6.8"
   s.add_dependency "blacklight-gallery", "0.3.0"
 
   ##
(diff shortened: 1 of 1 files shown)


no-reply:
Because `krikri` is a library (an engine), rather than an application, I think we want to keep the dependencies as loose as is practical. Being to prescriptive will force us to keep the dependencies in sync across applications.

AudreyAltman:
@no-reply ok, do you want to just leave it the way it is, or is it worth pinning to minor versions?
And should we be declaring these gems in heidrun with pins to tiny versions, to make sure the dependencies stay tight?

AudreyAltman:
Per discussion with tech team, we are _not_ pinning gems to tiny versions by default in Krikri.  Gems can be pinned to tiny versions in Heidrun instead (there will be a PR to effect later).  The new draft of this PR pins gems to _minor_ versions, only if they were not already pinned to minor or tiny versions.  Those gems that were previously pinned to tiny versions remain so, as there may have good reason for specifying tiny versions.

krikri.gemspec
```diff
@@ -57,12 +57,12 @@ Gem::Specification.new do |s|
...
-  s.add_development_dependency 'pry-rails'
-  s.add_development_dependency 'timecop'
+  s.add_development_dependency "krikri-spec", "~> 0.0"
+  s.add_development_dependency "sqlite3", "~> 1.3"
+  s.add_development_dependency "jettywrapper", "~> 2.0"
+  s.add_development_dependency "rspec-rails", "~> 3.4"
+  s.add_development_dependency "webmock", "~> 2.1"
+  s.add_development_dependency "factory_girl_rails", "~> 4.4.0"
+  s.add_development_dependency "pry-rails", "~> 0.3"
+  s.add_development_dependency "timecop"
```
markbreedlove: Can you verify that `timecop` is safe to let upgrade indefinitely?

krikri.gemspec
```diff
@@ -57,12 +57,12 @@ Gem::Specification.new do |s|
...
 
-  s.add_development_dependency 'krikri-spec', '~> 0.0'
-  s.add_development_dependency "sqlite3"
-  s.add_development_dependency "jettywrapper", '~> 2.0'
-  s.add_development_dependency "rspec-rails", '~> 3.3'
-  s.add_development_dependency 'webmock'
-  s.add_development_dependency 'factory_girl_rails', '~>4.4.0'
-  s.add_development_dependency 'pry-rails'
-  s.add_development_dependency 'timecop'
+  s.add_development_dependency "krikri-spec", "~> 0.0"
```
markbreedlove: Is `~> 0.0` going to restrict `krikri-spec` in any way?

krikri.gemspec
```diff
@@ -22,25 +22,25 @@ Gem::Specification.new do |s|
...
+  s.add_dependency "jsonpath", '~> 0.5'
+  s.add_dependency "devise", "~> 3.4.1"
+  s.add_dependency "resque", "~> 1.26"
   s.add_dependency "analysand", "4.0.0"
-  s.add_dependency "yajl-ruby"
-  s.add_dependency "elasticsearch", "~>0.4.0"
-  s.add_dependency "nokogiri", ">=1.6.8"
+  s.add_dependency "yajl-ruby", "1.2"
+  s.add_dependency "elasticsearch", "~> 0.4.0"
+  s.add_dependency "nokogiri", ">= 1.6.8"
```
markbreedlove: Are we confident that there won't be any breakage if `nokogiri` is upgraded to new major versions?

krikri.gemspec
```diff
@@ -22,25 +22,25 @@ Gem::Specification.new do |s|
...
+  s.add_dependency "audumbla", "~> 0.2"
   s.add_dependency "rdf-turtle", "~>1.1.8"
   s.add_dependency "rdf", "~> 1.1.13"
   s.add_dependency "dpla-map", "4.0.0.0.pre.13"
-  s.add_dependency "rest-client"
-  s.add_dependency "rdf-marmotta", '>= 0.0.6'
-  s.add_dependency "blacklight", "~>5.8.0"
-  s.add_dependency "therubyracer"
+  s.add_dependency "rest-client", "~> 2.0"
+  s.add_dependency "rdf-marmotta", ">= 0.0.6"
```
markbreedlove: Are we sure that `rdf-marmotta` won't have breaking changes between major versions?

markbreedlove:
BTW, @AudreyAltman, I think the Travis build error will be fixed after you rebase on `develop`.

krikri.gemspec
AudreyAltman: I missed timecop!  My bad, I'll pin it.

krikri.gemspec
AudreyAltman: Good catch, I'll change it from `>=` to `~>`, which will still ensure that the version is at least 0.0.6.  Same for nokogiri (in response to your comment below).

krikri.gemspec
AudreyAltman: @markbreedlove I don't know how else we would specify it, since v0.1 hasn't come out yet.  So far the one and only version is 0.0.1.  Maybe there's less need to worry about pinning since we manage `krikri-spec` directly in relation to `krikri`.  Any thoughts on how to better specify the dependency?

markbreedlove: I _think_ it should be `~> 0.0.1` so that it goes `0.0.2` ... `0.0.3` but not `0.1.0`.

I believe that the convention is that, for major version 0, which is assumed to be immature, you upgrade the tiny version always and treat it as if it were the minor version; and upping the minor version is saying "there may be breaking changes".

markbreedlove: Someone else can comment to verify ... @anarchivist ?

I'm being lazy :-) ... Have been googling and haven't found confirmation of this yet.

AudreyAltman: Huh, I did not know that.  Thanks!  I'll make the change.

AudreyAltman:
@markbreedlove okay, all your comments should have been addressed in the latest push.

krikri.gemspec
AudreyAltman: I likewise have so far been unable to google my way to confirmation ;)

markbreedlove:
Thanks, @AudreyAltman !  Looks good and appears to deploy well, so I'll merge it.

PR #266, opened 2016-09-07
DSL method for accessing field names

no-reply:
The new `#field_names` method returns a value array with the available field names (labels of child nodes) for the current node. This supports cases like `{'language' => { 'english' => 'some value', 'french' => 'some other value' }` with `.language.field_names` giving the keys.

### DIFF
--- a/lib/krikri/parser.rb
+++ b/lib/krikri/parser.rb
@@ -301,6 +301,19 @@ module Krikri
       end
 
       ##
+      # Lists field names on this node.
+      #
+      # This can be useful in cases where the desired value is a key or child 
+      # node label, rather than a value.
+      #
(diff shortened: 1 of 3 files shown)


no-reply:
The simplest motivating case is: https://github.com/dpla/heidrun-mappings/compare/loc-tj#diff-1f3b357ed970d13a025d4ea38227555eR83

anarchivist:
:+1:

markbreedlove:
Hi, @no-reply. Can you elaborate more on the motivating case referenced above? I don't understand yet why we need to query the list of available field names for that example.

no-reply:
@markbreedlove  sorry, see line 86 in language. I might have just linked the wrong line.

markbreedlove:
OK!

PR #267, opened 2016-09-07
Handle `JSONParser#get_child_node` for nested arrays

no-reply:


anarchivist:
The changes are 👍 , but I'm wondering if there are any additional tests that are worth adding here. Can you also provide a little context about why the fix was needed?

no-reply:
This was admittedly a bit rushed, as I was feeling progressively worse yesterday (hence the lack of tests). The short version is that 'location' fields have a nested list [[{...}]]. The existing code couldn't handle this, as it would try to pass the field name to an array and fail on "No implicit cast String to Integer".

The change to #value is due to ChildDeclaration using it for comparison. nil would always match nil, leading to duplicate values. 

There are probably improvements that could be made, and I'm happy to work on tests. Whether we want to wait, or push forward and clean up later is another question.

markbreedlove:
Sent @no-reply a Slack message about this and its relation to an issue we were troubleshooting last month.

markbreedlove:
@no-reply @anarchivist @moltude: I am taking over the fixes to this branch in order to get this merged today.

markbreedlove:
sorry @no-reply @anarchivist @moltude, please ignore that last comment. I'm just going to merge this after rebasing.

markbreedlove:
I am merging this with an additional commit that temporarily works around an issue with Travis. Please see the comment in https://github.com/dpla/KriKri/pull/267/commits/25288e3b0fa5be95140039c5dbcc16a7a42c5ebf

PR #269, opened 2016-09-12
Update engine_cart to ~> 1.0.1 (fixes #8556, #8557)

anarchivist:
Supersedes #268.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ tomcat
 marmotta
 
 ./spec/internal
+.internal_test_app
 dump.rdb
 
 coverage
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ addons:
   code_climate:
     repo_token: 9b8c58a49bdfba5582d6a766fbe198e3720158e5e65c2ae00b1667e5e1bb8769
 env: CI=yes
-before_install: 'gem install bundler -v "~> 1.11.0"'
+before_install: gem install bundler -v "~> 1.11"
 matrix:
   allow_failures:
     - rvm: 2.1.6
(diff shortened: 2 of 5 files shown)


anarchivist:
Also pins `sass-rails` to `~> 4.0.5` to address a build failure seen here: https://travis-ci.org/dpla/KriKri/jobs/159334607

moltude:
Mark, I probably missing something stupid here but I'm not able to get this to generate. Just walking through the instructions for doing a clean rebuild.

```
bundle update
rake engine_cart:clean
rake engine_cart:generate
```
but when I run 
`rake db:migrate` throws an error 

```../krikri/app/models/krikri/user.rb:6:in `<module:Krikri>': uninitialized constant User (NameError)
	from ../krikri/app/models/krikri/user.rb:2:in `<top (required)>'```


This error doesn't seem connected to your PR so I probably have some local issues that need to be addressed and someone else might be a in better position to review this PR in the very short term.

anarchivist:
@moltude Do you still have a `spec/internal` directory?

moltude:
@anarchivist I do

anarchivist:
@moltude The more recent versions of  `engine_cart` change the generated app to be built in `.internal_test_app`. Can you try changing the directory and running `bundle exec rake db:migrate`?

moltude:
I got this to work against a fresh clone. Tentatively 👍  Anyone else want to review before merging ?

markbreedlove:
Looks fine to me. The lines that were copied from Engine Cart are quite wide, and we have tried elsewhere try to stick with and 80-character limit.  I'm not sure how common this is in Ruby apps, but there's a lot of logic now in the Gemfile, and it would be nice if a file specifying dependencies could be more data-driven. I know there may be nothing to be done about this here, and that it's Ruby code and is supposed to be evaluated as such; just observing.

moltude:
@anarchivist We should also update the documentation that refers to `spec/internal`

markbreedlove:
@anarchivist I think the CI tests aren't running because the last commit says `[ci skip]`

PR #270, opened 2016-09-15
add two field value reports

AudreyAltman:
This adds two new field value reports.  It has been tested locally, and addresses [ticket #8559](https://issues.dp.la/issues/8559).


moltude:
Looks good @AudreyAltman! 👍

